PR #51, opened 2022-06-08
feat: expressions as type parameters

### LINKED ISSUE #28: Expressions as type parameters?
**Is your feature request related to a problem? Please describe.**
Example use cases:
* Type parameter isTrained
* Number of instances

**Describe the solution you'd like**
In addition to type parameters that are types, we could have type parameters that are expressions. We should probably introduce different kinds for type parameters if we do this (hard-coded, not extensible):
* `CallableType` (or `$CallableType`) (**probably not needed**)
* `ExpressionType` (or `$ExpressionType`) (**new**)
  * `IntType `(or `$IntType`)
  * `FloatType` (or `$FloatType`)
  * `BooleanType` (or `$BooleanType`)
  * `StringType` (or `$StringType`)
* `NamedType` (or `$NamedType`)
* `UnionType` (or `$UnionType`) (**probably not needed**)

Possible notation for kinds:
```
Class C<T::IntType>
```
or we reuse the `:` we already have for types (a kind being a type of types).

**Additional context (optional)**
* Type parameters are meant as parameters of types rather than parameters that are types.
* Without some prefix like `expr` this will only work for a subset of expressions (references + member accesses are problematic since they clash with named types and member types).
* At the very least we should offer literals + arithmetic & logical operations
* Relates to lars-reimann/safe-data-science#37 (which could be expressed by adding a SchemaType)
* Set of kinds can be hard-coded.
* Also: **optional** type parameters?

MShahzaib:
Closes #28

Summary of Changes
Added the following expressions that can be used as type parameters.

$SchemaType
$ExpressionType
$IntType
$FloatType
$BooleanType
$StringType
$NamedType

### DIFF
--- a/DSL/de.unibonn.simpleml/src/main/kotlin/de/unibonn/simpleml/validation/declarations/TypeParameterChecker.kt
+++ b/DSL/de.unibonn.simpleml/src/main/kotlin/de/unibonn/simpleml/validation/declarations/TypeParameterChecker.kt
@@ -1,0 +1,25 @@
+package de.unibonn.simpleml.validation.declarations
+
+import de.unibonn.simpleml.constant.SmlKind
+import de.unibonn.simpleml.constant.SmlVariance
+import de.unibonn.simpleml.constant.kind
+import de.unibonn.simpleml.constant.variance
+import de.unibonn.simpleml.simpleML.SimpleMLPackage.Literals
+import de.unibonn.simpleml.simpleML.SmlTypeParameter
+import de.unibonn.simpleml.validation.AbstractSimpleMLChecker
+import de.unibonn.simpleml.validation.codes.ErrorCode
+import org.eclipse.xtext.validation.Check
+
+class TypeParameterChecker : AbstractSimpleMLChecker() {
+
+    @Check
+    fun nonStaticPropagates(smlTypeParameter: SmlTypeParameter) {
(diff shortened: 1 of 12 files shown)


DSL/de.unibonn.simpleml/src/main/kotlin/de/unibonn/simpleml/validation/declarations/TypeParameterChecker.kt
```diff
@@ -0,0 +1,25 @@
...
+import de.unibonn.simpleml.simpleML.SimpleMLPackage.Literals
+import de.unibonn.simpleml.simpleML.SmlTypeParameter
+import de.unibonn.simpleml.validation.AbstractSimpleMLChecker
+import de.unibonn.simpleml.validation.codes.ErrorCode
+import org.eclipse.xtext.validation.Check
+
+class TypeParameterChecker : AbstractSimpleMLChecker() {
+
+    @Check
+    fun nonStaticPropagates(smlTypeParameter: SmlTypeParameter) {
```
lars-reimann, line 16: Function should be renamed

lars-reimann (changes requested):
One minor change needed, the rest looks excellent

lars-reimann (approved):

PR #89, opened 2022-07-06
feat: add atomic schema effects

### LINKED ISSUE #81: Add atomic schema effects

Add atomic schema effects (goals) for allowing data schema validation of datasets

### Desired solution
Goals that can be used in Predicate or Constraint constructs like

$ReadSchema(datasetName, datasetPath)
$CheckColumn(schemaIn, columnName)

MShahzaib:
Closes #81.

### Summary of Changes

- Added atomic schema effects "$ReadSchema" and "$CheckColumn" to the grammar.
- Added type checking for arguments of schema effects.


DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/staticAnalysis/linking/ArgumentToParameter.kt
```diff
@@ -45,3 +48,37 @@ fun SdsArgumentList.parametersOrNull(): List<SdsParameter>? {
...
+/**
+ * Returns the list of [SdsParameter]s that corresponds to this list of [SdsGoalArgument]s or `null` if it cannot be
+ * resolved.
+ */
+fun SdsGoalArgumentList.parametersOrNull(): List<SdsParameter>? {
+    return when (val parent = this.eContainer()) {
+        is SdsGoalCall -> parent.parametersOrNull()
+        else -> null
+    }
+}
```
lars-reimann, line 84: Should be in a separate file

DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/validation/typeChecking/GoalArgumentTypeChecker.kt
```diff
@@ -0,0 +1,41 @@
...
+            var argumentTypeString = argumentType.toSimpleString()
+            var parameterTypeString = parameterType.toSimpleString()
+
+            if (argumentTypeString == parameterTypeString) {
+                argumentTypeString = argumentType.toString()
+                parameterTypeString = parameterType.toString()
+            }
+
+            error(
+                "An goal argument of type '$argumentTypeString' cannot be assigned to a parameter of type '$parameterTypeString'.",
```
lars-reimann, line 35:
```suggestion
                "A goal argument of type '$argumentTypeString' cannot be assigned to a parameter of type '$parameterTypeString'.",
```

DSL/com.larsreimann.safeds/model/SafeDS.ecore
```diff
@@ -194,6 +198,11 @@
         <eStructuralFeatures xsi:type="ecore:EReference" name="columnName" eType="#//SdsString" containment="true" />
         <eStructuralFeatures xsi:type="ecore:EReference" name="columnType" eType="#//SdsAbstractType" containment="true" />
     </eClassifiers>
+    <eClassifiers xsi:type="ecore:EClass" name="SdsSchemaEffectPredicate" eSuperTypes="#//SdsAbstractGoalCallable">
+        <eStructuralFeatures xsi:type="ecore:EAttribute" name="effect" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString" />
+        <eStructuralFeatures xsi:type="ecore:EReference" name="parameterList" eType="#//SdsParameterList" containment="true" />
+        <eStructuralFeatures xsi:type="ecore:EReference" name="resultList" eType="#//SdsResultList" containment="true" />
+    </eClassifiers>
```
lars-reimann, line 205: This does not appear in the grammar. What is its purpose?

DSL/com.larsreimann.safeds/model/SafeDS.ecore
```diff
@@ -155,6 +156,9 @@
     <eClassifiers xsi:type="ecore:EClass" name="SdsGoalReference" eSuperTypes="#//SdsAbstractGoalExpression">
         <eStructuralFeatures xsi:type="ecore:EReference" name="declaration" eType="#//SdsAbstractDeclaration" />
     </eClassifiers>
+    <eClassifiers xsi:type="ecore:EClass" name="SdsSchemaEffectReference" eSuperTypes="#//SdsAbstractGoalExpression">
```
lars-reimann, line 159: Since this PR is about atomic schema effect, I'd prefer calling this "SdsAtomicSchemaEffect".

DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/staticAnalysis/linking/ArgumentToParameter.kt
lars-reimann: Also: Is there a way to do without all this duplication? Maybe make `SdsArgument` a supertype of `SdsGoalArgument` and `SdsArgumentList` a supertype of `SdsGoalArgumentList`?  Or is the behavior of them somehow different?

DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/staticAnalysis/GoalCallToGoalCallablel.kt
```diff
@@ -0,0 +1,54 @@
+package com.larsreimann.safeds.staticAnalysis
```
lars-reimann, line 1: Is there a way to get rid of this duplication? Maybe make `SdsCall` a superclass of `SdsGoalCall`? Do all scoping rules and validations of `SdsCall` also apply to `SdsGoalCall`?

lars-reimann (changes requested):
Main objective would be to look for ways to reduce code duplication.

DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/validation/typeChecking/GoalArgumentTypeChecker.kt
MShahzaib: will correct.

DSL/com.larsreimann.safeds/model/SafeDS.ecore
MShahzaib: SdsSchemaEffectPrediate nodes are created for schema atomic effects, the only reason is to have a node with parameter list and result list just like predicate would have so the argument type checking and later result type inference can be done the same way as for predicates. Think of them  as built-in predicates that the user can not make.
see  staticAnalysis/schema/SchemaEffectPredicate.kt

lars-reimann: Wouldn't it make more sense to just create those classes by hand in Kotlin? Personally, I like to avoid Ecore as much as possible.

lars-reimann: Some more thoughts:

We can also add a validator just for `SdsSchemaEffectReference`, which checks the following:

- `SdsSchemaEffectReference` must be called
- `SdsSchemaEffectReference` must have m-n arguments (in case some are optional)
- `SdsSchemaEffectReference` must have arguments with the correct types

MShahzaib: I removed schemaEffectPredicate.

Currently, my understanding It that named type parameters won't work for the atomicSchemaEffects. is that correct?

DSL/com.larsreimann.safeds/model/SafeDS.ecore
MShahzaib: Done !

DSL/com.larsreimann.safeds/src/main/kotlin/com/larsreimann/safeds/staticAnalysis/GoalCallToGoalCallablel.kt
MShahzaib: we need to have a short meeting to discuss this, I am not sure how to do this or even if we should.

DSL/com.larsreimann.safeds/model/SafeDS.ecore
lars-reimann: Can you give an example what you mean by "named type parameters"?

MShahzaib: sorry I meant named arguments. Instead of positional arguments, you can refer to the parameter name and pass arguments irrespective of the position in the paramterlist.

`$ReadSchema(datasetName="aName", datasetPath="aPath")`
and
`$ReadSchema(datasetPath="aPath", datasetName="aName")`
both are valid

lars-reimann: Would be good for consistency, if named arguments would work in this case.

MShahzaib (commented):
I did the changes

lars-reimann:
Please fix the errors induced by incorrect merging. Then I'll review this PR.

lars-reimann (approved):

PR #329, opened 2023-01-20
feat: Added method sort_columns in Table and changed equals to not care about order

### LINKED ISSUE #305: The equals method in Table (and Row) should not care about the order of the columns

If the order of columns is changed but the content of the table stays the same the equals method from Table returns `false`.

### Desired solution

The order of the columns should not matter. It would be good to add a method `sort_columns(lambda: Callable[[Column, Column], int]): Table` in Table to reorder the columns and use this function in the equals method. Default is alphabetical column names.

You could also just compare the Tables without considering the order of the columns

dorianheidorn:
Closes #305 .

### Summary of Changes

Added method sort_columns in Table and changed equals to not care about order of columns. The sort_columns method by default sorts after alphabetically order which can be changed by porividing a lambda-function as parameter.

### DIFF
--- a/Runtime/safe-ds/safe_ds/data/_table.py
+++ b/Runtime/safe-ds/safe_ds/data/_table.py
@@ -620,7 +620,10 @@ class Table:
 
     def sort_columns(
         self,
-        query: Callable[[Column, Column], int] = lambda col1, col2: (col1.name > col2.name) - (col1.name < col2.name),
+        query: Callable[[Column, Column], int] = lambda col1, col2: (
+            col1.name > col2.name
+        )
+        - (col1.name < col2.name),
     ) -> Table:
         """
         Sort a Table with the given lambda function.
@@ -651,10 +654,7 @@ class Table:
             return True
         table1 = self.sort_columns()
         table2 = other.sort_columns()
-        return (
-            table1._data.equals(table2._data)
-            and table1.schema == table2.schema
-        )
+        return table1._data.equals(table2._data) and table1.schema == table2.schema
 
     def __hash__(self) -> int:
         return hash(self._data)
--- a/Runtime/safe-ds/safe_ds/data/_table.py
+++ b/Runtime/safe-ds/safe_ds/data/_table.py
@@ -206,7 +206,8 @@ class Table:
         data_to_csv.to_csv(path_to_file, index=False)
 
     def rename_column(self, old_name: str, new_name: str) -> Table:
-        """Rename a single column by providing the previous name and the future name of it.
+        """
+        Rename a single column by providing the previous name and the future name of it.
 
         Parameters
         ----------
@@ -239,7 +240,8 @@ class Table:
         return Table(new_df.rename(columns={old_name: new_name}))
 
     def get_column(self, column_name: str) -> Column:
-        """Returns a new instance of Column with the data of the described column of the Table.
+        """
+        Returns a new instance of Column with the data of the described column of the Table.
 
         Parameters
         ----------
@@ -267,7 +269,8 @@ class Table:
         raise UnknownColumnNameError([column_name])
 
     def drop_columns(self, column_names: list[str]) -> Table:
-        """Returns a Table without the given columns
+        """
+        Returns a Table without the given columns
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


WinPlay02 (changes requested):
Blocked

WinPlay02 (approved):

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -612,12 +618,43 @@ def list_columns_with_non_numerical_values(self) -> list[Column]:
...
+        self,
+        query: Callable[[Column, Column], int] = lambda col1, col2: (
+            col1.name > col2.name
+        )
+        - (col1.name < col2.name),
+    ) -> Table:
+        """
+        Sort a Table with the given lambda function.
+        If no function is given the columns will be sorted alphabetically.
+        This function uses bubble sort.
```
WinPlay02, line 631:
```suggestion
        This function uses the default python sort algorithm.
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -612,12 +618,43 @@ def list_columns_with_non_numerical_values(self) -> list[Column]:
...
 
+    def sort_columns(
+        self,
+        query: Callable[[Column, Column], int] = lambda col1, col2: (
+            col1.name > col2.name
+        )
+        - (col1.name < col2.name),
+    ) -> Table:
+        """
+        Sort a Table with the given lambda function.
```
WinPlay02, line 629:
```suggestion
        Sort a table with the given lambda function.
```

WinPlay02 (commented):
Please update docs

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -612,12 +618,43 @@ def list_columns_with_non_numerical_values(self) -> list[Column]:
...
+        If no function is given the columns will be sorted alphabetically.
+        This function uses bubble sort.
+        The query should return:
+            TRUE if both columns should be switched
+            FALSE if the columns should not be switched
```
WinPlay02, line 634:
```suggestion
        The query should return:
            0, if both columns are equal
            < 0, if the first column should be ordered after the second column
            > 0, if the first column should be ordered before the second column
```

Gerhardsa0 (dismissed):

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -267,7 +270,8 @@ def get_column(self, column_name: str) -> Column:
         raise UnknownColumnNameError([column_name])
 
     def drop_columns(self, column_names: list[str]) -> Table:
-        """Returns a Table without the given columns
+        """
+        Returns a Table without the given columns
```
WinPlay02, line 274:
```suggestion
        Returns a table without the given columns
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -300,7 +304,8 @@ def drop_columns(self, column_names: list[str]) -> Table:
         return Table(transformed_data)
 
     def keep_columns(self, column_names: list[str]) -> Table:
-        """Returns a Table with exactly the given columns
+        """
+        Returns a Table with exactly the given columns
```
WinPlay02, line 308:
```suggestion
        Returns a table with exactly the given columns
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -346,7 +351,8 @@ def to_rows(self) -> list[Row]:
         ]
 
     def filter_rows(self, query: Callable[[Row], bool]) -> Table:
-        """Returns a Table with rows filtered by applied lambda function
+        """
+        Returns a Table with rows filtered by applied lambda function
```
WinPlay02, line 355:
```suggestion
        Returns a table with rows filtered by applied lambda function
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -239,7 +241,8 @@ def rename_column(self, old_name: str, new_name: str) -> Table:
         return Table(new_df.rename(columns={old_name: new_name}))
 
     def get_column(self, column_name: str) -> Column:
-        """Returns a new instance of Column with the data of the described column of the Table.
+        """
+        Returns a new instance of column with the data of the described column of the Table.
```
WinPlay02, line 245:
```suggestion
        Returns a new instance of column with the data of the described column of the table.
```

WinPlay02 (approved):

WinPlay02 (approved):

PR #332, opened 2023-01-20
feat: Added ability to provide a TableSchema in the Table constructor

### LINKED ISSUE #306: Create empty table with initiated scheme

When working with Tables, it is not possible to create an empty table with a presetted schema. This is a Problem if the empty table should be used to be filled while working with it because every operation like "add_row()". In this case the schema of both entities, the table and row are not matching.
For further information on why there are mismatching schemas please refer to #296 

### Desired solution

Let schema of an empty table be initially None and change other methods for this functionality.

```py
table = Table([])
table.add_row(Row([1,2])
```

dorianheidorn:
Closes #306 .

### Summary of Changes

Added ability to provide a TableSchema in the Table constructor

### DIFF
--- a/Runtime/safe-ds/safe_ds/data/_table.py
+++ b/Runtime/safe-ds/safe_ds/data/_table.py
@@ -33,9 +33,9 @@ class Table:
         data: typing.Iterable
             the data you want to save in the table
         schema: TableSchema
-            the Schema for the table
-            None be default, if None the Schema will be provided by the data that is given
-            If a Schema is set there is no check if this schema is compatible with the data!
+            the schema for the table
+            None by default, if None the schema will be provided by the data that is given
+            If a schema is set, there is no check if this schema is compatible with the data!
         """
         self._data: pd.Dataframe = (
             data if isinstance(data, pd.DataFrame) else pd.DataFrame(data)
--- a/Runtime/safe-ds/safe_ds/data/_table.py
+++ b/Runtime/safe-ds/safe_ds/data/_table.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import os.path
 import typing
 from pathlib import Path
-from typing import Callable, Union, Optional
+from typing import Callable, Optional, Union
 
 import pandas as pd
 from pandas import DataFrame, Series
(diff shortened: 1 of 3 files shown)


Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -23,11 +24,29 @@
...
+        """
+        Create a new Table
+
+        Parameters
+        ----------
+        data: typing.Iterable
+            the data you want to save in the table
+        schema: TableSchema
+            the Schema for the table
+            None be default, if None the Schema will be provided by the data that is given
```
WinPlay02, line 37:
```suggestion
            None by default, if None the Schema will be provided by the data that is given
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -23,11 +24,29 @@
...
+            If a Schema is set there is no check if this schema is compatible with the data!
```
WinPlay02, line 38:
```suggestion
            If a schema is set, there is no check if this schema is compatible with the data!
```

Runtime/safe-ds/safe_ds/exceptions/_data_exceptions.py
```diff
@@ -81,6 +81,15 @@ def __init__(self) -> None:
         super().__init__("Failed because at least two schemas didn't match.")
 
 
+class MissingSchemaError(Exception):
+    """
+    Exception raised when a required Schema is missing.
```
WinPlay02, line 86:
```suggestion
    Exception raised when a required schema is missing.
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -23,11 +24,29 @@
...
+    def __init__(self, data: typing.Iterable, schema: Optional[TableSchema] = None):
+        """
+        Create a new Table
+
+        Parameters
+        ----------
+        data: typing.Iterable
+            the data you want to save in the table
+        schema: TableSchema
+            the Schema for the table
```
WinPlay02, line 36:
```suggestion
            the schema for the table
```

Gerhardsa0 (approved):

PR #351, opened 2023-01-27
feat: remove_outliers() method for table class

### LINKED ISSUE #308: Remove Outliers method for Table class

In many cases it's useful to remove extreme values from your dataset. Currently there is no way to achieve this.

### Desired solution

Add a `Table.remove_outliers()` method that removes every row that is more then 3 standard deviations removed from the average.

Marvjowa:
Closes #308.

### Summary of Changes

Adds Table.remove_outliers() method that removes every row that is more then 3 standard deviations removed from the average. Also adds Table.list_columns_with_numerical_values(), which is used in Table.remove_outliers(). Also tests.

### Testing Instructions

Create Table, call Table.remove_outliers(), check if expected rows have been removed.


Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -705,6 +722,45 @@ def sort_columns(
...
+            a new table where the outliers have been removed
+        """
+        result = self._data.copy(deep=True)
+
+        table_without_nonnumericals = Table.from_columns(
+            self.list_columns_with_numerical_values()
+        )
+
+        print(table_without_nonnumericals._data)
+        print(np.absolute(stats.zscore(table_without_nonnumericals._data)))
```
WinPlay02, line 742:
```suggestion
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -641,6 +643,21 @@ def list_columns_with_non_numerical_values(self) -> list[Column]:
                 cols.append(self.get_column(column_name))
         return cols
 
+    def list_columns_with_numerical_values(self) -> list[Column]:
+        """
+        Get a list of Columns only containing numerical values
```
WinPlay02, line 648:
```suggestion
        Get a list of columns only containing numerical values
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -641,6 +643,21 @@ def list_columns_with_non_numerical_values(self) -> list[Column]:
...
+
+        Returns
+        -------
+        cols: list[Column]
+            the list with only numerical Columns
```
WinPlay02, line 653:
```suggestion
            the list with only numerical columns
```

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -705,6 +722,45 @@ def sort_columns(
         columns.sort(key=functools.cmp_to_key(query))
         return Table.from_columns(columns)
 
+    def remove_outliers(self) -> Table:
+        """
+        Removes all rows from the table that contain at least one outlier defined as having a value that is more than 3
+        standard deviations removed from the column average.
```
WinPlay02, line 728:
```suggestion
        Removes all rows from the table that contain at least one outlier defined as having a value that deviates more than 3
        from the column average.
```

WinPlay02 (commented):
Please consider suggestions

WinPlay02 (approved):

WinPlay02 (approved):

PR #359, opened 2023-01-27
fix: 352 some methods dont work with empty tables

### LINKED ISSUE #352: Some methods don't work with empty tables
### Describe the bug

Curently a user is supposed to create a empty table like this:
```
table_schema = # initialize TableSchema
table = Table([], table_schema)
```
For some methods like get_column, we check the TableSchema if the respective column exists. Then we try to slice it from the Dataframe. The problem is, that this column doesn't exists, so we get a error in the iloc() method call.

### To Reproduce

Create a empty Table and call it's get_column() method on a column that should exists but is empty.

### Expected behavior

The Dataframe should always match the TableSchema, so that the checks we do are valid. 

### Additional context (optional)

In the tests for Table.remove_outliers() and Table.list_columns_with_numerical_values() a termorary fix was applied so that another pull request could get closed. These need to be deleted.
Additionally we should add more tests for empty tables, to ensure, that this won't be an issue in the future.

dorianheidorn:
Closes #352 .

### Summary of Changes

Added functionality to correctly add a Table without data but correct and checkable schema.
Also added a test to verify correct checkability of empty created table


WinPlay02 (dismissed):

Runtime/safe-ds/safe_ds/data/_table.py
```diff
@@ -284,12 +286,14 @@ def get_column(self, column_name: str) -> Column:
...
         if self.schema.has_column(column_name):
-            return Column(
+            output_col = Column(
                 self._data.iloc[
                     :, [self.schema._get_column_index_by_name(column_name)]
                 ].squeeze(),
                 column_name,
             )
+            output_col._type = self.schema.get_type_of_column(column_name)
+            return output_col
```
GideonKoenig, line 296:
```suggestion
            output_column = Column(
                self._data.iloc[
                    :, [self.schema._get_column_index_by_name(column_name)]
                ].squeeze(),
                column_name,
            )
            output_column._type = self.schema.get_type_of_column(column_name)
            return output_column
```

Marvjowa (approved):

GideonKoenig (approved):

Marvjowa (approved):

PR #360, opened 2023-01-27
feat: OrdinalEncoder

### LINKED ISSUE #307: OrdinalEncoder

Values that can be ordered in a useful way should be encoded differently than values which do not have an inherent order.

### Desired solution

An OrdinalEncoder should be added to the `safe_ds.data` package. The order needs to be provided in the fit method (and which Columns should be encoded as optional).

Use the LabelEncoder or OneHotEncoder to encode these values. The downside of this approach is, that the information about the order is lost.

### Additional context

It would be useful to implement Issue #318 before this one

WinPlay02:
Closes #307 

### Summary of Changes

Added Ordinal Encoder
Added __eq__ method to TableSchema -> fixed Table __eq__
Double checked the label encoder, one hot encoder
Added tests


Marsmaennchen221 (approved):

Runtime/safe-ds/safe_ds/data/_ordinal_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+        p_df = table._data
+        p_df.columns = table.schema.get_column_names()
+        try:
+            self.oe.fit(p_df[[column_name]])
+        except exceptions.NotFittedError as exc:
+            raise exceptions.LearningError("") from exc
+
+    def transform(self, table: Table, column_name: str) -> Table:
+        """
+        Transform the given Table to a ordinal-encoded table.
```
SmiteDeluxe, line 53:
```suggestion
        Transform the given Table to an ordinal-encoded table.
```

Runtime/safe-ds/safe_ds/data/_ordinal_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+         Raises
+         ------
+             a NotFittedError if the Model wasn't fitted before transforming
+        """
+        p_df = table._data.copy()
+        p_df.columns = table.schema.get_column_names()
+        try:
+            p_df[[column_name]] = self.oe.transform(p_df[[column_name]])
+            p_df[column_name] = p_df[column_name].astype(dtype="int64", copy=False)
+            return Table(p_df)
```
SmiteDeluxe, line 74:
```suggestion
            return Table(p_df, table.schema)
```
Otherwise columnnames wouldn't be ratained, no?

Runtime/safe-ds/safe_ds/data/_ordinal_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+        except Exception as exc:
+            raise exceptions.NotFittedError from exc
+
+    def fit_transform(self, table: Table, columns: list[str]) -> Table:
+        """
+        oridnal-encodes a given table with the given ordinal encoder
```
SmiteDeluxe, line 80:
```suggestion
        Oridnal-encodes a given table with the given ordinal encoder.
```

Runtime/safe-ds/safe_ds/data/_ordinal_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+        Raises
+        -------
+            NotFittedError if the encoder wasn't fitted before transforming.
+        """
+
+        p_df = table._data.copy()
+        p_df.columns = table.schema.get_column_names()
+        try:
+            p_df[[column_name]] = self.oe.inverse_transform(p_df[[column_name]])
+            return Table(p_df)
```
SmiteDeluxe, line 130:
```suggestion
            return Table(p_df, table.schema)
```
Same as above

SmiteDeluxe (changes requested):
Additional:
Missing the types of Return objects in Docstrings

SmiteDeluxe (approved):

PR #373, opened 2023-01-27
ci: disable markdownlint

lars-reimann:


WinPlay02 (dismissed):

lars-reimann:
Disabled in `https://github.com/lars-reimann/.github/blob/main/.mega-linter.yml` instead.

PR #396, opened 2023-01-31
fix: now mode returns a list

### LINKED ISSUE #395: mode: return a list with all values that occur as often as the most common value

Gerhardsa0:
Closes #395 .

### Summary of Changes

now mode returns a list of the most common elements

changed in imputer that the first element of the list is taken but maybe we want to change this

adjusted tests and fixed test_summary


Gerhardsa0:
@lars-reimann right now the Imputer just takes the first elements of the mode list, should we change it?
I am  not sure what behavior is wanted so I am asking here.

WinPlay02 (approved):

Issue #542, opened 2023-08-04
Port generator to Langium


lars-reimann:
Additional material:

* https://www.typefox.io/blog/code-generation-for-langium-based-dsls
* https://langium.org/tutorials/generation/

Tasks:
- [x] basic generator
- [x] remove varargs from tests
- [ ] filter Statements without effect (related to #603)
- [x] wildcard imports
- [x] verify special names when importing from pipeline/test inside the same package (see test generation/imports)

PR #634, opened 2023-10-12
feat: generation

WinPlay02:
Closes partially #542 

### Summary of Changes

- added an implementation of the generator
- enabled tests for the generator (and removed the dummy test)
- fixed an error related to constant expressions of integers being interpreted as floats
- fixed an error where test files were read in an incorrect order
- added a generation test for lists and maps
- removed varargs from existing tests


src/cli/generator.ts
```diff
@@ -1,41 +1,632 @@
...
+const getPythonNameOrDefault = function (
+    services: SafeDsServices,
+    object: SdsPipeline | SdsSegment | SdsParameter | SdsDeclaration,
+) {
+    return services.builtins.Annotations.getPythonName(object) || object.name;
+};
+
+const generateAssignee = function (assignee: SdsAssignee): string {
+    if (isSdsBlockLambdaResult(assignee)) {
+        return (assignee as SdsBlockLambdaResult).name;
```
lars-reimann, line 154: The explicit cast is not needed. Since the `isSdsXY` functions [return a type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), TypeScript can figure out that `assignee` is an `SdsBlockLambdaResult` inside this branch.

src/cli/generator.ts
```diff
@@ -1,41 +1,632 @@
...
+const generateModule = function (services: SafeDsServices, module: SdsModule): string {
+    const importSet = new Map<String, ImportData>();
+    const segments = streamAllContents(module)
+        .filter(isSdsSegment)
+        .map((segment) => generateSegment(services, segment, importSet))
+        .toArray();
+    const pipelines = streamAllContents(module)
+        .filter(isSdsPipeline)
+        .map((pipeline) => generatePipeline(services, pipeline, importSet))
+        .toArray();
```
lars-reimann, line 553: Segments and pipelines can only be module members, so you don't need to loop over all AST nodes in a module:

```suggestion
    const segments = moduleMembersOrEmpty(module)
        .filter(isSdsSegment)
        .map((segment) => generateSegment(services, segment, importSet))
        .toArray();
    const pipelines = moduleMembersOrEmpty(module)
        .filter(isSdsPipeline)
        .map((pipeline) => generatePipeline(services, pipeline, importSet))
        .toArray();
```

src/cli/generator.ts
```diff
@@ -1,41 +1,632 @@
...
+                    localValue.alias !== undefined
+                        ? `${localValue.declarationName} as ${localValue.alias}`
+                        : <string>localValue.declarationName,
+                ) || [];
+        declaredImports.push(`from ${key} import ${[...new Set(importedDecls)].join(', ')}`);
+    }
+    return [...new Set(qualifiedImports), ...new Set(declaredImports)];
+};
+
+const generateModule = function (services: SafeDsServices, module: SdsModule): string {
```
lars-reimann, line 544: Regarding the order of functions, please order them, so the calling functions are above the called functions where possible. This makes it easier to understand the program when reading it from top to bottom.

tests/language/generation/creator.ts
```diff
@@ -30,8 +30,13 @@ const createGenerationTest = async (parentDirectory: URI, inputUris: URI[]): Pro
     const actualOutputRoot = URI.file(path.join(parentDirectory.fsPath, 'generated'));
     const expectedOutputFiles = readExpectedOutputFiles(expectedOutputRoot, actualOutputRoot);
     let runUntil: Location | undefined;
+    // First read all stubs, then read all the other files; This should avoid broken references
+    const sortedInputUris = inputUris
```
WinPlay02, line 34: @lars-reimann The solution of calling `loadAllDocuments` before the call to `getErrors(services, code)` broke with duplicate definitions of segments and functions.
I temporarily used the old solution of sorting the files by extension. This could break, if definitions are needed from a file which is not an `.sdsstub` file.
Should I create a function similar to `getErrors(services, code)` below that supports working on a `LangiumDocument` instead of raw code?

lars-reimann: > Should I create a function similar to getErrors(services, code) below that supports working on a LangiumDocument instead of raw code?

That seems reasonable. It could also accept a URI instead of the document.

WinPlay02:
@lars-reimann There are still some uncovered lines or branches left. They are fallback code (e.g. throwing exceptions for unknown nodes) and should not be hit in the normal code generation flow, and therefore can not be tested with generation tests.
What would you suggest, to solve that problem?

lars-reimann:
> @lars-reimann There are still some uncovered lines or branches left. They are fallback code (e.g. throwing exceptions for unknown nodes) and should not be hit in the normal code generation flow, and therefore can not be tested with generation tests. What would you suggest, to solve that problem?

Add comments to those lines to selectively ignore them for coverage computation:

```ts
/* c8 ignore next */
const a = 1;

/* c8 ignore next 2 */
const b = 2;
const c = 3;

/* c8 ignore start */
if (false) {
    // ...
}
/* c8 ignore stop*/
```

lars-reimann (approved):
Great work!

PR #641, opened 2023-10-16
feat: check `@PythonName` and `@PythonModule`

lars-reimann:
Closes #XYZ


lars-reimann:
@WinPlay02 You can now get the Python names and Python modules via the methods `getPythonName` and `getPythonModule` of `services.builtins.Annotations`.

PR #684, opened 2023-10-23
feat: PythonCall annotation

### LINKED ISSUE #617: @PythonCall annotation for functions

For some functions, we might want to offer a different interface than Python offers. For example, we might want to replace the Python code `', '.join([1, 2])` with `[1, 2].join(', ')`.

### Desired solution

An annotation `@PythonCall` that is evaluated by the code generator that specifies how code should be rewritten. The annotation takes a string that can have placeholder that point either to `this` or a parameter. Example:

```
class List {
    @PythonCall("$delimiter.join($this)")
    fun join(delimiter: String) -> String
}
```

The code generator can then replace the placeholders with the actual values of `this` and the parameters. For the call `[1, 2].join(', ')`, it would then generate the Python code `', '.join([1, 2])`.

_No response_

### Screenshots (optional)

_No response_

_No response_

WinPlay02:
Closes #617 

### Summary of Changes

- added `PythonCall` annotation to builtins
- added code generation for direct function calls and member access function calls
- added tests

### DIFF
--- a/src/cli/generator.ts
+++ b/src/cli/generator.ts
@@ -389,11 +389,21 @@ const generateExpression = function (expression: SdsExpression, frame: Generatio
                 return generatePythonCall(pythonCall, argumentsMap);
             }
         }
-        if (isSdsMemberAccess(expression.receiver) && isSdsReference(expression.receiver.member) && isSdsFunction(expression.receiver.member.target.ref)) {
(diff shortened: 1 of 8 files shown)


lars-reimann (approved):
That was fast, thanks a lot!

PR #714, opened 2023-10-31
feat: Source Maps for Code Generation

### LINKED ISSUE #30: Generate source map
**Is your feature request related to a problem? Please describe.**
Generate a source map so we can link runtime errors in the Python code to DSL code.

See https://github.com/langium/langium/blob/7f3fc78a9fcedd92fe7cb848278e0fd5e8f88781/packages/langium/CHANGELOG.md#generator-tracing

WinPlay02:
Closes #30

### Summary of Changes

- changed code generation to return traced nodes instead of text
- add option to generate source maps by providing an optional path


WinPlay02:
@lars-reimann 
How should source maps be tested? Should they be added to the generation tests as an additional file to be generated?

lars-reimann:
> @lars-reimann How should source maps be tested? Should they be added to the generation tests as an additional file to be generated?

That seems reasonable.

lars-reimann:
TypeFox also published a series of blog posts about the code generation with Langium:

* https://www.typefox.io/blog/code-generation-for-langium-based-dsls/
* https://www.typefox.io/blog/code-generation-for-langium-based-dsls-2/
* https://www.typefox.io/blog/code-generation-for-langium-based-dsls-3/

The third one specifically talks about source maps.

packages/safe-ds-lang/tests/resources/generation/declarations/empty pipeline/output/sourcemap.json
lars-reimann, line 1: I'd suggest following this name convention for source maps:

> Optionally, a source map will have the same name as the generated file but with a “.map” extension.  For example, for “page.js” a source map named “page.js.map” would be generated. 

[Source](https://sourcemaps.info/spec.html)

packages/safe-ds-lang/tests/helpers/testResources.ts
```diff
@@ -59,13 +59,13 @@ export const listTestSafeDsFiles = (rootTestResourceName: TestResourceName): URI
...
- * Lists all Python files in the given root directory.
+ * Lists all Python files and JSON files in the given root directory.
  *
  * @param rootTestResourceName The resource name of the root directory.
  * @return URIs of the discovered Python files.
  */
-export const listTestPythonFiles = (rootTestResourceName: TestResourceName): URI[] => {
-    const pattern = `**/*.py`;
+export const listTestPythonJsonFiles = (rootTestResourceName: TestResourceName): URI[] => {
+    const pattern = `**/*.{json,py}`;
```
lars-reimann, line 68:
```suggestion
export const listTestFilesWithExtensions = (rootTestResourceName: TestResourceName, extensions: string[]): URI[] => {
    const pattern = `**/*.{${extensions.join}}`;
```

I think this more general function is probably easier to understand. The docstring also needs to be adjusted. You can leave the function for Safe-DS files, but maybe call `listTestFilesWithExtensions` internally.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -100,7 +111,7 @@ export class SafeDsPythonGenerator {
         this.partialEvaluator = services.evaluation.PartialEvaluator;
     }
 
-    generate(document: LangiumDocument, destination: URI): TextDocument[] {
+    generate(document: LangiumDocument, destination: URI, sourceMapDestination: URI | undefined): TextDocument[] {
```
lars-reimann, line 114: As explained below, there is a name convention for source maps. Thus, setting the destination of the source map is not needed. However, it would be nice to have a flag to configure whether a source map should be generated or now. 

Since JS/TS does not support named arguments directly, the usual approach is to wrap such options into an object:

```ts
interface GenerateOptions {
    destination: URI,
    createSourceMaps: boolean
}
```

This clarifies the meaning of the passed arguments:

```
generate(document, { destination: ..., createSourceMaps: true })
```

lars-reimann (changes requested):

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
lars-reimann: This would also allow the creation of source maps for multiple generated files in the same directory.

WinPlay02: Do you mean, the `GenerateOptions` interface should be moved from `cli/generate.ts` to `language/generation/safe-ds-python-generator.ts`?

lars-reimann: If the options are identical, yes.

lars-reimann: And talking about the CLI: A new flag for the `generate` command to control the creation of source maps from there would be nice, too.

WinPlay02: Other than the `destination` field being optional, they would be identical

packages/safe-ds-lang/tests/helpers/testResources.test.ts
```diff
@@ -44,7 +44,7 @@ describe('listTestSafeDsFiles', () => {
 describe('listTestPythonFiles', () => {
```
lars-reimann, line 44:
```suggestion
describe('listTestFilesWithExtensions', () => {
```

lars-reimann (changes requested):
Looks great, just one tiny change left.

lars-reimann (approved):
Thanks a lot!

PR #784, opened 2023-11-20
feat: purity computer

lars-reimann:
### Summary of Changes

Add a new service that infers whether a call or callable is pure or has side effects.


lars-reimann:
@WinPlay02 [services.purity.PurityComputer](https://github.com/Safe-DS/DSL/blob/9da61eede2386070854cf72a28d7245b54174a40/packages/safe-ds-lang/src/language/purity/safe-ds-purity-computer.ts) adds the methods you need to implement the remainder of #542. Let me know if something is missing.

PR #807, opened 2023-12-05
fix: Wait for tree-kill during deactivation if child process does not respond

WinPlay02:
If the child process does not respond, tree kill does not wait for the process to end.
This may cause VS Code to continue with shutting down, killing the first process of the tree only.
The reference to the tree is then lost, and tree-kill can no longer continue killing the other processes.
To fix this, await the completion of tree-kill.

Added some more logging

### DIFF
--- a/packages/safe-ds-vscode/src/extension/pythonServer.ts
+++ b/packages/safe-ds-vscode/src/extension/pythonServer.ts
@@ -61,12 +61,24 @@ export const startPythonServer = async function (): Promise<void> {
 };
 
 export const stopPythonServer = async function (): Promise<void> {
+    logOutput('Stopping python server...');
     if (pythonServer !== undefined) {
         if (
             (pythonServerAcceptsConnections && !(await requestGracefulShutdown(2500))) ||
             !pythonServerAcceptsConnections
         ) {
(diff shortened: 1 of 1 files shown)


packages/safe-ds-vscode/src/extension/pythonServer.ts
```diff
@@ -60,20 +60,37 @@ export const startPythonServer = async function (): Promise<void> {
...
             !pythonServerAcceptsConnections
         ) {
-            treeKill(pythonServer.pid!);
+            logOutput(`Tree-killing python server process ${pythonServer.pid}...`);
+            const pid = pythonServer.pid!;
+            // Wait for tree-kill to finish killing the tree
+            await new Promise<void>((resolve, _reject) => {
+                treeKill(pid, (error) => {
+                    resolve();
+                    if (error !== undefined && error !== null) {
```
lars-reimann, line 80:
```suggestion
                    if (error) {
```

It's not exactly the same, since it would not trigger if `error` is `false` or `""`, for example. But this is the usual way to check whether an object is defined.

lars-reimann (approved):

PR #821, opened 2024-01-09
feat: Partial execution of pipelines

WinPlay02:
- pipelines can now be partially executed by providing the target placeholder to calculate
- `executePipeline` now accepts an optional target placeholder name
- added a test for partial execution and all possible code dependencies
- fixed block lambdas not generating when used inside of an assignment


WinPlay02:
Should some utility functions moved to other files? (like `hasImpurityReasonEffectOnFuture`, `getAssignmentForPlaceholder`, `getPlaceholderByName` ...)

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +395,238 @@ export class SafeDsPythonGenerator {
...
+    private getAssignmentForPlaceholder(placeholder: SdsPlaceholder): SdsAssignment | undefined {
+        let assignment = placeholder.$container;
+        while (assignment !== undefined && !isSdsAssignment(assignment)) {
+            assignment = assignment.$container;
+        }
+        if (!assignment || !assignment.expression || !isSdsAssignment(assignment)) {
+            /* c8 ignore next 2 */
+            return undefined;
+        }
+        return assignment;
```
lars-reimann, line 483: Can be shortened by using `getContainerOfType(placeholder, isSdsAssignment)`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +395,238 @@ export class SafeDsPythonGenerator {
         )!;
     }
 
+    private getPlaceholderByName(block: SdsBlock, name: string): SdsPlaceholder | undefined {
+        return streamPlaceholders(block).find((placeholder) => placeholder.name === name);
+    }
```
lars-reimann, line 400: Could be useful elsewhere, so move to `nodeProperties.ts` (before/after `streamPlaceholders`).

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +395,238 @@ export class SafeDsPythonGenerator {
...
+        }
+        // Endless recursions don't have any effect on others
+        if (reasonPast === EndlessRecursion || reasonFuture === EndlessRecursion) {
+            /* c8 ignore next 2 */
+            return false;
+        }
+        // For OtherImpurityReason, PotentiallyImpureParameterCall, UnknownCallableCall we can't say.
+        // So it might have an effect on other impure functions
+        return true;
+    }
```
lars-reimann, line 607: We could replace the if/else chain by polymorphism, i.e. add a method to `ImpurityReason` whether it affects another, later impurity reason.

packages/safe-ds-vscode/src/extension/pythonServer.ts
```diff
@@ -244,8 +244,14 @@ export const tryMapToSafeDSSource = async function (
  * @param services SafeDsServices object, used to import the pipeline file.
  * @param pipelinePath Path to a Safe-DS pipeline file to execute.
  * @param id A unique id that is used in further communication with this pipeline.
+ * @param targetPlaceholder The name of the target placeholder, used to do partial execution.
```
lars-reimann, line 247: Also explain that if it's not specified, everything is run.

lars-reimann:
I'll do a more thorough review later.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+                referencedStatement,
+                allStatements,
+                neededPlaceholders,
+                neededStatements,
+            );
+        }
+    }
+
+    private isNodeInsideBlockLambda(node: AstNode): boolean {
+        return Boolean(getContainerOfType(node, isSdsBlockLambda));
```
lars-reimann, line 463: Use `hasContainerOfType`. Then it's also doubtful whether this extra method is even need or whether it can be inlined.

packages/safe-ds-lang/tests/resources/generation/partial/input.sdstest
lars-reimann, line 1: While this looks very comprehensive, it would be nice if you could split this into separate test cases. Currently, it's difficult to tell which aspect of the `run until` logic is tested where.

packages/safe-ds-lang/tests/resources/generation/expressions/block lambda/input.sdstest
lars-reimann, line 1: What is this change about?

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -368,7 +375,13 @@ export class SafeDsPythonGenerator {
         frame: GenerationInfoFrame,
         generateLambda: boolean = false,
     ): CompositeGeneratorNode {
+        let targetPlaceholder = frame.targetPlaceholder
+            ? getPlaceholderByName(block, frame.targetPlaceholder)
+            : undefined;
```
lars-reimann, line 380: Allow the second parameter of `getPlaceholderByName` to be `undefined` and return `undefined` in that case. This makes it easier for the caller, since they can avoid conditionals. We've frequently used this pattern for other methods.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+        neededPlaceholders: Set<string>,
+        neededStatements: Set<SdsStatement>,
+    ) {
+        // If placeholder is inside a lambda, do not continue; Only placeholders inside the pipeline are collected
+        if (this.isNodeInsideBlockLambda(targetPlaceholder)) {
+            return;
+        }
+        // Find assignment of placeholder, to search used placeholders and impure dependencies
+
+        let assignment = getContainerOfType(targetPlaceholder, isSdsAssignment);
```
lars-reimann, line 425: Extraneous blank line.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+            /* c8 ignore next 2 */
+            return;
+        }
+        // If placeholder is already found, do not continue; It has been handled before
+        if (neededPlaceholders.has(targetPlaceholder.name) && neededStatements.has(assignment)) {
+            return;
+        }
+        neededPlaceholders.add(targetPlaceholder.name);
+        neededStatements.add(assignment);
+        // Get referenced placeholders of current placeholder. An expression in the assignment will always exist here
```
lars-reimann, line 436:
```suggestion
        // Get references to current placeholder. An expression in the assignment will always exist here
```

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+        let placeholders = streamAllContents(assignment.expression!)
```
lars-reimann, line 437: Rename to `referencedPlaceholders`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+        }
+        // Find assignment of placeholder, to search used placeholders and impure dependencies
+
+        let assignment = getContainerOfType(targetPlaceholder, isSdsAssignment);
+        if (!assignment || !assignment.expression) {
+            /* c8 ignore next 2 */
+            return;
+        }
+        // If placeholder is already found, do not continue; It has been handled before
+        if (neededPlaceholders.has(targetPlaceholder.name) && neededStatements.has(assignment)) {
```
lars-reimann, line 431: This seems to be the only time `neededPlaceholders` is ever read and for this check the second condition `neededStatements.has(assignment)` seems to be sufficient.

Unless, there's another reason to keep `neededPlaceholders` around, I'd suggest to
1. Remove the parameter `neededPlaceholders`.
2. Make the `neededStatements` parameter the return value.
3. Rename the method `getStatementsRelevantFor...` (or `keepStatementsRelevantFor`).

This applies to `getAllPlaceholdersAndStatementsRelevantForPlaceholder` and `getAllPlaceholdersAndStatementsRelevantForStatement`.

-----

Generally, the logic in and interaction between
* `getAllPlaceholdersAndStatementsRelevantForPlaceholder`,
* `getAllPlaceholdersAndStatementsRelevantForStatement`,
* `getImpureDependenciesForStatement`

seems somewhat convoluted.

Isn't it possible to
1. Get the assignment containing the target placeholder (we always must evaluate this one).
2. Collect all placeholder references & impurity reasons in its expression.
3. Iteratively traverse the previous statements in reverse order and for each
    a. Check whether it defines a needed placeholder or interact with some known impurity reason. 
    b. If neither applies, we can omit it and continue with the previous one.
    c. Otherwise, we must generate code for it and collect all placeholder references & impurity reasons in its expression.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
lars-reimann, line 425: Can be `const` (prefer `const` over `let`).

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
lars-reimann, line 437: Can be `const`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -381,20 +394,180 @@ export class SafeDsPythonGenerator {
...
+        for (const referencedPlaceholder of placeholders) {
+            this.getAllPlaceholdersAndStatementsRelevantForPlaceholder(
+                referencedPlaceholder,
+                allStatements,
+                neededPlaceholders,
+                neededStatements,
+            );
+        }
+        // Get impure dependencies and follow them
+        let impureDependencies = this.getImpureDependenciesForStatement(assignment, allStatements);
```
lars-reimann, line 451: Can be `const`. Please check later code as well, I'm not going to repeat this for every occurrence.

lars-reimann (changes requested):

lars-reimann (approved):
Fantastic work, thanks a lot!

PR #822, opened 2024-01-11
feat: pre-load all relevant workspace files before executing pipeline

WinPlay02:
- relevant workspace files are preloaded and refreshed on each execution
- `executePipeline` function now accepts a `LangiumDocument` instead of a pipeline path
-  fix: filling console causing unresponsiveness when handling new runner output by truncating too long output


packages/safe-ds-vscode/src/extension/pythonServer.ts
```diff
@@ -435,7 +406,11 @@ const connectToWebSocket = async function (): Promise<void> {
                     logOutput(`[Runner] Message received: (${event.type}, ${typeof event.data}) ${event.data}`);
                     return;
                 }
-                logOutput(`[Runner] Message received: '${event.data}'`);
+                logOutput(
```
WinPlay02, line 409: @SmiteDeluxe According to my tests, this change fixes the unresponsiveness after loading the large table. Please test, if this is the case for you.

packages/safe-ds-vscode/src/extension/mainClient.ts
```diff
@@ -128,29 +128,70 @@ const acceptRunRequests = function (context: vscode.ExtensionContext) {
...
-            }
-            if (!pipelinePath) {
-                vscode.window.showErrorMessage('Could not run Safe-DS Pipeline, as no pipeline is currently selected.');
-                return;
-            }
-            const pipelineId = crypto.randomUUID();
-            printOutputMessage(`Launching Pipeline (${pipelineId}): ${pipelinePath}`);
-            executePipeline(services, pipelinePath.fsPath, pipelineId);
-        }),
+        vscode.commands.registerCommand(
```
lars-reimann, line 131: The function `acceptRunRequests` gets quite long, and it does more than its name implies. Better split this into three separate ones:
1. Register message callbacks
2. Register VS Code commands (maybe even extract the callback for each command into a named function)
3. Register watchers

Later we can move parts of this into separate files.

packages/safe-ds-vscode/src/extension/mainClient.ts
```diff
@@ -128,29 +128,70 @@ const acceptRunRequests = function (context: vscode.ExtensionContext) {
...
+                const validationErrorMessage = await validateDocuments(services, unvalidatedSdsDocuments);
+                if (validationErrorMessage) {
+                    vscode.window.showErrorMessage(validationErrorMessage);
+                    return;
+                }
+                // Run it
+                const pipelineId = crypto.randomUUID();
+                printOutputMessage(`Launching Pipeline (${pipelineId}): ${pipelinePath}`);
+                let mainDocument;
+                if (!services.shared.workspace.LangiumDocuments.hasDocument(URI.parse(pipelinePath.toString()))) {
```
lars-reimann, line 182:
```suggestion
                if (!services.shared.workspace.LangiumDocuments.hasDocument(pipelinePath)) {
```

Doesn't this work too?

lars-reimann (changes requested):

lars-reimann (approved):
Looks good. Let's wait for @SmiteDeluxe's response before merging, though.

packages/safe-ds-vscode/src/extension/pythonServer.ts
lars-reimann: I'll merge it now, we've waited long enough for a response.

PR #826, opened 2024-01-18
feat: support placeholder queries that only request a subset of data

WinPlay02:
- changed `placeholder_query` messages (VSCode to Runner) to allow optional `window_begin` and `window_size` fields (in addition to the placeholder `name`)
- changed `placeholder_value` messages (Runner to VSCode) to include `window_begin`, `window_size` and `window_max` fields and `windowed` flag when querying only windows of data


packages/safe-ds-vscode/src/extension/messages.ts
```diff
@@ -68,7 +68,17 @@ export interface ProgramMainInformation {
 export interface PlaceholderQueryMessage {
     type: 'placeholder_query';
     id: string;
-    data: string;
+    data: PlaceholderQuery;
+}
+
+/**
+ * Contains the query for a placeholder value.
+ * Field name contains the name of the requested placeholder, field window_begin may contain an offset of and window_size may contain the size of a subset of the requested data.
```
lars-reimann, line 76: Better add a `/** ... */` comment to the individual fields.

packages/safe-ds-vscode/src/extension/messages.ts
```diff
@@ -68,7 +68,27 @@ export interface ProgramMainInformation {
...
+
+    /**
+     * The offset of the requested data.
+     */
+    window_begin?: number;
+
+    /**
+     * The size of the requested data.
+     */
+    window_size?: number;
```
lars-reimann, line 91: Is it intended that you can also specify only `window_begin` or `window_size`? What's the behavior in this case? I guess `window_begin` defaults to the start of the data and `window_size` to the size of the remaining data?

packages/safe-ds-vscode/src/extension/messages.ts
```diff
@@ -105,6 +125,10 @@ export interface PlaceholderValue {
     name: string;
     type: string;
     value: string;
+    windowed?: boolean;
+    window_begin?: number;
+    window_size?: number;
+    window_max?: number;
```
lars-reimann, line 131: I have a similar question here. Though the combination `windowed: false` and `window_size: 50` doesn't seem to make much sense either way. It seems like in this case, either all fields must be set or none. Thus, I'd propose to add some nesting here:

```ts
{
    // ...
    value: string;
    window?: {
        begin: number;
        size: number;
        max: number;
    }
}
```

For consistency, we should then also use this nested structure for the `PlaceholderQuery`.

lars-reimann (changes requested):

packages/safe-ds-vscode/src/extension/messages.ts
WinPlay02: That is correct

packages/safe-ds-vscode/src/extension/messages.ts
WinPlay02: Where applicable, comments for other messages were also restructured.

lars-reimann (approved):

PR #827, opened 2024-01-18
feat: Memoization

WinPlay02:
- memoizable function are now generated with `runner_memoized_function_call`
- hidden parameters are inserted, if needed


packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -634,27 +638,29 @@ export class SafeDsPythonGenerator {
...
+            // Memoize constructor or function call
+            if (isSdsFunction(callable) || isSdsClass(callable)) {
+                if (isSdsFunction(callable)) {
+                    const pythonCall = this.builtinAnnotations.getPythonCall(callable);
+                    if (pythonCall) {
+                        let thisParam: CompositeGeneratorNode | undefined = undefined;
+                        if (isSdsMemberAccess(expression.receiver)) {
+                            thisParam = this.generateExpression(expression.receiver.receiver, frame);
+                        }
+                        const argumentsMap = this.getArgumentsMap(expression.argumentList.arguments, frame);
```
lars-reimann, line 651: Should use `getArguments(expression)`. It's better to decouple the code as much as possible from the abstract syntax, so it's easier to change it.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+            expression,
+            frame,
+        )}", lambda *_ : ${generatedPythonCall}, [${joinTracedToNode(expression.argumentList, 'arguments')(
+            memoizedArgs,
+            (arg) => this.generateExpression(arg, frame),
+            { separator: ', ' },
+        )}], [${joinToNode(hiddenParameters, (param) => param, { separator: ', ' })}])`;
+    }
+
+    private isCallMemoizable(expression: SdsCall): boolean {
```
lars-reimann, line 818: `isMemoizableCall` would be more consistent with the other methods like `generateMemoizedCall`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+        const impurityReasons = this.purityComputer.getImpurityReasonsForExpression(expression);
+        let memoizable = true;
+        for (const reason of impurityReasons) {
+            // If the file is not known, the call is not memoizable
+            if (!(reason instanceof FileRead) || reason.path === undefined) {
+                memoizable &&= false;
+            }
+        }
+        return memoizable;
```
lars-reimann, line 827: Can be expressed with `some` or `every`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+        frame.addImport({ importPath: RUNNER_SERVER_PIPELINE_MANAGER_PACKAGE });
+        const hiddenParameters = this.getMemoizedCallHiddenParameters(expression, frame);
+        const memoizedArgs = getParameters(this.nodeMapper.callToCallable(expression)).map(
+            (parameter) => this.nodeMapper.callToParameterValue(expression, parameter)!,
+        );
+        const containsOptionalArgs =
+            sortedArgs
+                .map((arg) => this.nodeMapper.argumentToParameter(arg))
+                .filter((param) => param)
+                .find((param) => !Parameter.isRequired(param)) !== undefined;
```
lars-reimann, line 845: Can use `some`.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+            // Verify same package
+            if (value.package !== targetModule.name) {
+                continue;
+            }
+            if (isSdsQualifiedImport(value)) {
+                const importedDeclarations = getImportedDeclarations(value);
+                for (const importedDeclaration of importedDeclarations) {
+                    if (declaration === importedDeclaration.declaration?.ref) {
+                        return `${
+                            this.builtinAnnotations.getPythonModule(targetModule) || value.package
```
lars-reimann, line 923: Abstract this into its own method, akin to `getPythonNameOrDefault` (`getPythonModuleOrDefault`?).

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+                            frame,
+                        )})`,
+                    );
+                }
+            }
+        }
+        return hiddenParameters;
+    }
+
+    private getFullyQualifiedReferenceName(expression: SdsExpression, declaration: SdsDeclaration): string | undefined {
```
lars-reimann, line 904: I don't understand this method. 

You always return the equivalent of
`this.builtinAnnotations.getPythonModule(targetModule) || targetModule.name}.${this.getPythonNameOrDefault(declaration)`, unless the declaration is in a different module and not imported explicitly. In the latter case you return undefined.

Why does it make a difference whether the declaration is imported explicitly or not? For instance, everything in a `safeds.xy` package is available without imports, but the notion of a qualified name still makes sense.

At first glance it seems like you can remove this method completely and use `getQualifiedName` instead.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +793,168 @@ export class SafeDsPythonGenerator {
...
+                return `${
+                    this.builtinAnnotations.getPythonModule(targetModule) || value.package
+                }.${this.getPythonNameOrDefault(declaration)}`;
+            }
+        }
+        /* c8 ignore next */
+        return undefined;
+    }
+
+    private generateFullyQualifiedFunctionName(
```
lars-reimann, line 938: This seems brittle, since you build the qualified name based on the call and not the callee. There might be multiple ways to reach the same method, however. So, you are better off getting the (unique) callee first, and then getting its qualified name.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
WinPlay02: Thanks for the hint, that makes it a lot easier. I cannot use `getQualifiedName` directly because of python module/python name differences.

WinPlay02:
Maybe it makes sense to move `getPythonModuleOrDefault`, `getPythonNameOrDefault` and `getQualifiedNamePythonCompatible` to other files, as they might be useful for other purposes?

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
lars-reimann: Right, that makes sense.

lars-reimann:
Currently, I don't see other places than the code generator that would need to process `@PythonName` and `@PythonModule`. If that changes, we can move them to a common location.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -775,6 +816,116 @@ export class SafeDsPythonGenerator {
...
+    ): CompositeGeneratorNode {
+        frame.addImport({ importPath: RUNNER_SERVER_PIPELINE_MANAGER_PACKAGE });
+        const hiddenParameters = this.getMemoizedCallHiddenParameters(expression, frame);
+        const memoizedArgs = getParameters(this.nodeMapper.callToCallable(expression)).map(
+            (parameter) => this.nodeMapper.callToParameterValue(expression, parameter)!,
+        );
+        const containsOptionalArgs = sortedArgs
+            .map((arg) => this.nodeMapper.argumentToParameter(arg))
+            .filter((param) => param)
+            .some((param) => Parameter.isOptional(param));
```
lars-reimann, line 860: `some` can replace the entire `map`/`filter`/`some` chain:

```ts
sortedArgs.some(arg => {
    const param = this.nodeMapper.argumentToParameter(arg));
    return Parameter.isOptional(param); // returns false if param is undefined
})
```

lars-reimann (changes requested):
Just one small suggestion left.

PR #832, opened 2024-01-25
fix: generation of static class functions

WinPlay02:
- Fixes static functions being generated with `this` parameter, which would be the class itself
- added test for this case


lars-reimann (approved):
Thanks for the fix.

PR #836, opened 2024-01-31
feat: optionally generate code without runner integration

### LINKED ISSUE #831: Basic Python code generation

The current code generator creates code that can only be executed with the [Safe-DS/Runner](https://github.com/Safe-DS/Runner). It would be nice if it could also generate code that can be run with a normal Python interpreter that has the required DS packages.

### Desired solution

Add a flag to the code generator (off by default). If it is set, the code generator no longer creates code to save placeholders or to memoize functions.

Open question: How do we handle `eager_or`, `eager_and`, `eager_elvis`, `safe_access`? To preserve the same semantics, we might just add the complete functions to the generated files, as needed, rather than an import. Maybe we always do this and remove this functionality from the runner, for simplicity? We could add the `CODEGEN_PREFIX` to the function names to prevent collisions with user-defined declarations.

_No response_

### Screenshots (optional)

_No response_

_No response_

WinPlay02:
Closes #831

- allow not generating runner function calls (memoized function call, placeholder saving)
- added test for eject functionality + tests that begin with "eject" are automatically tested without runner support


packages/safe-ds-lang/tests/resources/generation/eject/output/tests/generator/eject/gen_input.py
```diff
@@ -0,0 +1,8 @@
+# Pipelines --------------------------------------------------------------------
+
+def test():
+    f((g() or g()))
+    f((g() and g()))
```
lars-reimann, line 5: These are not equivalent to the runner calls, since the right operand of `or`/`and` no longer get executed even if the left operand determines the result. The Safe-DS `or`/`and` is deliberately not short-circuited.

You could instead generate another immediately-invoked lambda.

lars-reimann: Another option would be to add the functions that are currently in `runner.codegen` to the generated code and call those. This would be more readable than the lambda, IMO, since the function has a proper name. Since this feature is used when users want to edit the Python code afterward, the generated code should be as readable as possible.

I'd imagine something like this:

```py
# Utils ------------------------------------------------------------------------

def __gen_eager_or(left_operand: bool, right_operand: bool) -> bool:
    return left_operand or right_operand


def __gen_eager_and(left_operand: bool, right_operand: bool) -> bool:
    return left_operand and right_operand

# Pipelines --------------------------------------------------------------------

def test():
    f(__gen_eager_or(g(), g()))
    f(__gen_eager_and(g(), g()))
```

The utils should only be added when they are needed. We can then always generate this code, regardless of the `disableRunnerIntegration` flag and remove the `src/safeds_runner/codegen/_operators.py` file in the runner.

In that case, the difference between the two generation modes is simply whether calls get memoized and placeholders get saved.

packages/safe-ds-lang/tests/language/generation/safe-ds-python-generator.test.ts
```diff
@@ -36,6 +36,7 @@ describe('generation', async () => {
                     destination: test.actualOutputRoot,
                     createSourceMaps: true,
                     targetPlaceholder: runUntilPlaceholderName,
+                    disableRunnerIntegration: test.testName.startsWith('[eject'), // Tests in the "eject" suite are tested with disabled runner integration
```
lars-reimann, line 39: The `creator.ts` files are meant to create full descriptions of the tests. The `xyz.test.ts` then simply run those tests. In light of this separation of concerns, I'd suggest updating the `generation/creator.ts` file instead:

* Add a `disableRunnerIntegration` property to the `GenerationTest` interface.
* Populate the property in `createGenerationTest` by checking the file name.
* In `safe-ds-python-generator.test.ts`, simply pass the value of this property along.

lars-reimann: Also, update `docs/development/generation-testing.md`, please.

lars-reimann (changes requested):

packages/safe-ds-lang/tests/language/generation/safe-ds-python-generator.test.ts
lars-reimann, line 39: Also, update `docs/development/generation-testing.md`, please.

docs/development/generation-testing.md
```diff
@@ -12,6 +12,10 @@ generation test.
 
         If you want to skip a test, add the prefix `skip-` to the folder name.
 
+    !!! tip "Tests without runner integration"
+
+        If you want to create a test without runner integration (memoization and placeholder saving), put it in the eject folder or use eject as a preix for a new top level folder.
```
lars-reimann:
```suggestion
        If you want to create a test without runner integration (memoization and placeholder saving), put it in the `eject` folder or use `eject` as a prefix for a new top level folder.
```

docs/development/generation-testing.md
```diff
@@ -12,6 +12,10 @@ generation test.
 
         If you want to skip a test, add the prefix `skip-` to the folder name.
 
+    !!! tip "Tests without runner integration"
+
+        If you want to create a test without runner integration (memoization and placeholder saving), put it in the `eject` folder or use `eject` as a preix for a new top level folder.
```
WinPlay02, line 17:
```suggestion
        If you want to create a test without runner integration (memoization and placeholder saving), put it in the `eject` folder or use `eject` as a prefix for a new top level folder.
```

lars-reimann (approved):
Looks great. 🚀

PR #837, opened 2024-01-31
docs: describe runner installation and configuration process

WinPlay02:


WinPlay02:
Is there any reason why the image file is not found?

lars-reimann (approved):
Great input, thank you.

PR #888, opened 2024-02-16
feat: check the runner version to be compatible

### LINKED ISSUE #880: Check whether runner version is compatible

Not all versions of the DSL and the runner are compatible to each other.

### Desired solution

* When starting the runner from the DSL, check whether the installed version is compatible. The package [semver](https://www.npmjs.com/package/semver) might be helpful here (`satisfies` method).
* Update the README to include the supported runner version (rather than saying to download the latest one). Ideally, the installation command should already contain the version bound.

_No response_

### Screenshots (optional)

_No response_

_No response_

WinPlay02:
### Summary of Changes

- check the runner version to be compatible before starting the runner

Closes #880


lars-reimann (approved):
👍

PR #914, opened 2024-02-22
fix: runner startup silent failure

### LINKED ISSUE #909: No feedback if runner command is wrong
### Describe the bug

If a runner with an incorrect version is installed, a popup is shown to the user. If the extension is started with an incorrect runner command, however, no popup is displayed and nothing is logged. Moreover, changes to the runner command have no effect until the extension is restarted.

### To Reproduce

1. Set `safe-ds-runner-test` as the runner command.
2. Restart the extension host.
3. Open some Safe-DS file.

### Expected behavior

1. A popup should be shown.
2. Something should be logged.
3. Changes to the runner command should lead to another attempt to start it.

### Screenshots (optional)

_No response_

_No response_

WinPlay02:
Closes #909

### Summary of Changes

- failing to start runner process now logs an appropriate message informs the user via popup again
- command is now successfully passed to the runner component, if changed

### DIFF
--- a/packages/safe-ds-lang/src/language/runner/safe-ds-runner.ts
+++ b/packages/safe-ds-lang/src/language/runner/safe-ds-runner.ts
@@ -105,7 +105,9 @@ export class SafeDsRunner {
             }
         } catch (error) {
             this.logging.outputError(`Could not start runner: ${error instanceof Error ? error.message : error}`);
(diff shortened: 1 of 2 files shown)


lars-reimann (approved):
Works perfectly, thanks for the fast fix.

PR #946, opened 2024-03-07
refactor: pass pipeline name for runner execution

WinPlay02:
- pipeline name is now passed to execution to allow execution of multiple pipelines

### DIFF
--- a/packages/safe-ds-lang/src/language/runner/safe-ds-runner.ts
+++ b/packages/safe-ds-lang/src/language/runner/safe-ds-runner.ts
@@ -15,8 +15,7 @@ import { BasicSourceMapConsumer, SourceMapConsumer } from 'source-map';
 import treeKill from 'tree-kill';
 import { SafeDsAnnotations } from '../builtins/safe-ds-annotations.js';
 import { SafeDsPythonGenerator } from '../generation/safe-ds-python-generator.js';
-import { isSdsModule, isSdsPipeline } from '../generated/ast.js';
-import { getModuleMembers } from '../helpers/nodeProperties.js';
+import { isSdsModule } from '../generated/ast.js';
 import semver from 'semver';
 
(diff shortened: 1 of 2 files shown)


WinPlay02:
@SmiteDeluxe @lars-reimann The pipeline name is now needed when execution a pipeline.
I did not implement the proposed idea that combines the pipeline name with the target placeholder, as full executions would need another function.

packages/safe-ds-lang/src/language/runner/safe-ds-runner.ts
```diff
@@ -283,11 +282,13 @@ export class SafeDsRunner {
...
      * @param pipelineDocument Document containing the main Safe-DS pipeline to execute.
      * @param id A unique id that is used in further communication with this pipeline.
+     * @param pipelineName Name of the pipeline that should be run
      * @param targetPlaceholder The name of the target placeholder, used to do partial execution. If no value or undefined is provided, the entire pipeline is run.
      */
     public async executePipeline(
         pipelineDocument: LangiumDocument,
         id: string,
+        pipelineName: string,
         targetPlaceholder: string | undefined = undefined,
```
lars-reimann, line 292:
```suggestion
        id: string,
        pipelineDocument: LangiumDocument,
        pipelineName: string,
        targetPlaceholder: string | undefined = undefined,
```

I'd like to have the parameters for documents, pipeline and placeholder next to each other. Since `targetPlaceholder` is optional, this is the only way.

WinPlay02:
> > @SmiteDeluxe @lars-reimann The pipeline name is now needed when execution a pipeline. I did not implement the proposed idea that combines the pipeline name with the target placeholder, as full executions would need another function.
> 
> Your solution is cleaner anyway. Is this ready for review or do you still work on it?

It should now be ready

lars-reimann (approved):

PR #950, opened 2024-03-29
feat: stubs for `safe-ds` library

lars-reimann:
Closes #926

### Summary of Changes

Add a first version of stubs for the `safe-ds` library.

PR #954, opened 2024-03-30
feat: add code to pipelines (#929) & in the end full profiling and refactors/fixes

### LINKED ISSUE #929: EDA: Create logic for appending code to pipelines
1. Webview triggers for actions that manipulate data
2. VS Code to pick those up
3. Extend code to pipeline and execute
4. Get and send new placeholders to webview to update

SmiteDeluxe:
Closes #929 

### Summary of Changes

Communication from EDA -> Runner in own file that allows appending of code to the executed pipeline to get and generate new info. 
This results in full profiling being done already, fetched when needed and displayed in webview.
Also includes start of from webview triggered actions that append code, in the form of filters. But this is to be picked up in new issue.

Other changes include refactorings and fixes for visual and edge case bugs.

### Important to run this:
Some environments might need a change in the runner for images (plots) to work, that will be pushed to the Runner repo soon.
To PipelineManager:
```
from safeds.data.image.containers import Image
```
To beginning of save_placeholder method:
```
import torch
        if isinstance(value, Image):
            value = Image(value._image_tensor, torch.device("cpu"))
```


SmiteDeluxe:
@GideonKoenig @WinPlay02 Only added you as Reviewers if you want to, but probably not needed. In any case appreciated.

General context is I have to get this through rather quickly so I can continue working on things building on top of this.

WinPlay02 (commented):
During testing, I got this error message after clicking on "Show Profiling", which I cannot reproduce anymore:

```
rejected promise not handled within 1 second: RangeError: Invalid array length
extensionHostProcess.js:144
stack trace: RangeError: Invalid array length
	at SafeDsPythonGenerator.generateAssignment (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69167:102)
	at SafeDsPythonGenerator.generateStatement (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69138:33)
	at joinTracedToNode.separator (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69095:78)
	at -\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68679:152
	at -\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68658:21
	at reduceWithIsLast (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68689:14)
	at joinToNode (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68654:10)
	at -\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68679:43
	at SafeDsPythonGenerator.generateBlock (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69095:50)
	at SafeDsPythonGenerator.generatePipeline (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69067:170)
	at -\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69001:94
	at Array.map (<anonymous>)
	at SafeDsPythonGenerator.generateModule (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:69001:71)
	at SafeDsPythonGenerator.generate (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:68916:34)
	at SafeDsRunner.generateCodeForRunner (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:77436:47)
	at SafeDsRunner.executePipeline (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:77416:50)
	at -\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:79203:42
	at new Promise (<anonymous>)
	at RunnerApi.addToAndExecutePipeline (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:79180:12)
	at RunnerApi.getProfiling (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:79341:16)
	at _EDAPanel.createOrShow (-\DSL\packages\safe-ds-vscode\dist\extension\mainClient.cjs:79596:84)
	at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

I sadly cannot view the profiling information, maybe it's something wrong with my setup this time? It just stays at the "Loading..." state

lars-reimann:
I'll review this tomorrow.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+            return 'var(--warn-color)';
+        } else if (profilingItem.interpretation === 'bold') {
+            return 'var(--font-dark)';
+        } else if (profilingItem.interpretation === 'default') {
+            return 'var(--font-light)';
+        } else if (profilingItem.interpretation === 'category') {
+            return 'var(--font-light)';
+        } else {
+            return 'var(--font-light)';
+        }
```
lars-reimann, line 22: A `switch` would reduce noise here, since all branches immediately return a value, so no `break` is needed.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+    };
+</script>
+
+<div class="wrapper">
+    <div class="profilingItemsTop">
+        {#each profiling.top as profilingTopItem}
```
lars-reimann, line 28: What is the _semantic_ difference between `profiling.top` and `profiling.bottom`? Better use that as the name of those properties in the model. How this is then displayed is dictated by the Svelte components.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+            {#if profilingTopItem.type === 'text'}
+                <div class="profilingItem">
+                    <span style="color: {getProfilingItemColor(profilingTopItem)};">{profilingTopItem.value}</span>
```
lars-reimann, line 31: Where possible, keep styling out of JS. Here, you can use `profilingItem.interpretation` in a class attribute and set the colors via CSS.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+                </div>
+            {:else if profilingTopItem.type === 'numerical'}
+                <div class="profilingItem">
+                    <span class="profilingItemFirst" style="color: {getProfilingItemColor(profilingTopItem)};"
```
lars-reimann, line 35: Suggestion for class: `profilingItemKey` instead of `profilingItemFirst`. Otherwise, I'd expect this to be used for all first children of `profilingItem`, in which case it could also be replaced by a pseudoclass.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+                    >
+                    <span style="color: {getProfilingItemColor(profilingTopItem)};">{profilingTopItem.value}</span>
+                </div>
+            {:else if profilingTopItem.type === 'image'}
+                <div class="profilingItem">
+                    <img
+                        style="width: {imageWidth}px;"
+                        class="profilingImage"
+                        src={'data:image/' + profilingTopItem.value.format + ';base64,' + profilingTopItem.value.bytes}
+                        alt="profiling histogram"
```
lars-reimann, line 46: Is this always a histogram or should the alt text be part of the image type?

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+                        class="profilingImage"
+                        src={'data:image/' +
+                            profilingBottomItem.value.format +
+                            ';base64,' +
+                            profilingBottomItem.value.bytes}
+                        alt="profiling histogram"
+                    />
+                </div>
+            {/if}
+        {/each}
```
lars-reimann, line 80: Create a new component (e.g. `ProfilingItemList` or just `ProfilingItem` if you don't want to include the loop) to get rid of this lengthy duplicated code.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
```diff
@@ -0,0 +1,116 @@
...
+                    <span class="profilingItemFirst" style="color: {getProfilingItemColor(profilingBottomItem)};"
+                        >{profilingBottomItem.name}:</span
+                    >
+                    <span style="color: {getProfilingItemColor(profilingBottomItem)};">{profilingBottomItem.value}</span
+                    >
+                </div>
+            {:else if profilingBottomItem.type === 'image'}
+                <div class="profilingItem">
+                    <img
+                        style="width: {imageWidth}px;"
```
lars-reimann, line 70: Svelte has a shorthand syntax for styles: `style:width="..."` (https://svelte.dev/docs/element-directives#style-property).

packages/safe-ds-eda/src/webviewState.ts
```diff
@@ -26,6 +26,32 @@ window.addEventListener('message', (event) => {
...
+                currentState.update((state) => {
+                    return {
+                        ...state,
+                        table: {
+                            ...state.table!,
+                            columns: state.table!.columns.map((column) => {
+                                const profiling = message.value.find((p) => p.columnName === column[1].name);
+                                if (profiling) {
+                                    return [
+                                        column[0],
```
lars-reimann, line 40: Some thoughts/questions:

1. `column[0]`/`column[1]` reads like getting the first/second value of a column. However, IIUC, `column[0]` is some index and `column[1]` is the actual column. This naming is very confusing.
2. What is the purpose of `column[0]`? It seems to only ever be written but never read. Moreover, the actual column has a name, which we can use as an identifier.

packages/safe-ds-eda/src/webviewState.ts
```diff
@@ -26,6 +26,32 @@ window.addEventListener('message', (event) => {
...
             currentState.set(message.value);
             break;
+        case 'setProfiling':
+            if (get(currentState) && get(currentState).table) {
+                currentState.update((state) => {
+                    return {
+                        ...state,
+                        table: {
+                            ...state.table!,
+                            columns: state.table!.columns.map((column) => {
```
lars-reimann, line 36: A flatter state is easier to update and reason about. Here, you could store profiling information as a top-level item in the state as an object that maps column names to profiling information. Then you can quickly set the profiling information given a column name.

packages/safe-ds-eda/types/state.d.ts
```diff
@@ -77,26 +77,25 @@ export interface Profiling {
 
 export interface ProfilingDetailBase {
     type: 'numerical' | 'image' | 'name';
-    name: string;
+    value: string;
+    interpretation: 'warn' | 'error' | 'default' | 'bold' | 'good';
```
lars-reimann, line 81: Is `interpretation` relevant to all subtypes of `ProfilingDetailBase`, including `ProfilingDetailImage`? If not, better add an intermediate interface like `ProfilingDetailsText`.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
SmiteDeluxe: there is no real semantic difference, it is just what is displayed at top and what at the bottom, separated by the space in between, who's size is dictated by the biggest profiling info.

packages/safe-ds-eda/types/state.d.ts
```diff
@@ -77,26 +77,25 @@ export interface Profiling {
 
 export interface ProfilingDetailBase {
```
lars-reimann, line 78: Does this need to be exported? It seems like the subtypes + `ProfilingDetail` should be enough.

lars-reimann, line 78: What is the meaning of `bold`? The other words make more sense as an interpretation.

packages/safe-ds-eda/types/state.d.ts
```diff
@@ -148,11 +151,18 @@ export interface ValueRangeFilter extends ColumnFilterBase {
...
-export type CategoricalFilter = SearchStringFilter;
+export interface PossibleSpecificValueFilter extends ColumnFilterBase {
+    type: 'specificValue';
+    values: string[];
+}
+
+export type NumericalFilter = ValueRangeFilter;
+export type CategoricalFilter = SearchStringFilter | SpecificValueFilter;
+
+export type PossibleColumnFilter = ValueRangeFilter | PossibleSearchStringFilter | PossibleSpecificValueFilter;
```
lars-reimann, line 165: * For consistency, rename this to `ColumnFilter`. 
* Also, the interface `ColumnFilterBase` probably needn't be exported anymore.

packages/safe-ds-eda/types/state.d.ts
```diff
@@ -77,26 +77,25 @@ export interface Profiling {
 
 export interface ProfilingDetailBase {
     type: 'numerical' | 'image' | 'name';
```
lars-reimann, line 79: If you use discriminating unions (here the type `ProfilingDetails`), there's no reason to include the type field in a parent interface.

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
lars-reimann: It looks like `top` is always `[validRatio, missingRatio]`, which can be displayed regardless of the type of the column. It seems like the restriction "can be displayed regardless of type" could be used to differentiate what goes to the top and what goes to the bottom.

Thus, I'd suggest replacing e.g.

```ts
profiling: {
    top: [validRatio, missingRatio],
    bottom: [
        { type: 'text', value: 'Categorical', interpretation: 'bold' },
        ...uniqueProfilings,
    ],
},
```

with

```ts
profiling: {
    validRatio,
    missingRatio,
    other: [
        { type: 'text', value: 'Categorical', interpretation: 'bold' },
        ...uniqueProfilings,
    ],
},
```

This way, the model does not already dictate the representation.

lars-reimann:
I'll continue tomorrow with the last 4 files:

* TableView
* runnerApi
* edaPanel
* mainClient.

WinPlay02:
> @PythonCall("from safeds.data.tabular.containers import Column")
> fun __import_Column()

@SmiteDeluxe I've been using the officially provided stubs added in #950, and I removed all custom provisional stubs.

WinPlay02:
After comparing with your provided stub, it seems that calls may differ between yours and the official stubs.

E.g. 
- `plot_histogram` -> `plotHistogram` 
- `get_column` -> `getColumn`
- ...

lars-reimann:
> After comparing with your provided stub, it seems that calls may differ between yours and the official stubs.
> 
> E.g.
> 
>     * `plot_histogram` -> `plotHistogram`
> 
>     * `get_column` -> `getColumn`
> 
>     * ...

That's OK. All those functions have a `PythonName` annotation, to map our name convention to the Python name convention.

WinPlay02:
> > After comparing with your provided stub, it seems that calls may differ between yours and the official stubs.
> > E.g.
> > ```
> > * `plot_histogram` -> `plotHistogram`
> > 
> > * `get_column` -> `getColumn`
> > 
> > * ...
> > ```
> 
> That's OK. All those functions have a `PythonName` annotation, to map our name convention to the Python name convention. Example:
> 
> ```
> @Pure
> @PythonName("plot_histogram")
> fun plotHistogram() -> result1: Image
> ```
> 
> Inside `Safe-DS` this function has the name `plotHistogram`, but your code generator then translates this back to `plot_histogram`.

Yes, that's correct, but not what I meant.

To clarify: In the `runnerApi.ts`-file, calls are made to specifically `plot_histogram` (not `plotHistogram` as specified in the stub from #950).

As I understand, the code generator can't find the function (as `plot_histogram` does not exist in the stub) in this case and fails.

lars-reimann:
> As I understand, the code generator can't find the function (as `plot_histogram` does not exist in the stub) in this case and fails.

That is correct. When @SmiteDeluxe switches from his stubs to the included ones, the Safe-DS code must be adjusted.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+        newPlaceholderName: string,
+    ): string {
+        return (
+            'val ' +
+            newPlaceholderName +
+            ' = ' +
+            tablePlaceholder +
+            '.get_column("' +
+            columnName +
+            '").missing_value_ratio(); \n'
```
lars-reimann, line 91:
```suggestion
            '").missingValueRatio(); \n'
```

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+        columnName: string,
+        tablePlaceholder: string,
+        newPlaceholderName: string,
+    ): string {
+        return (
+            'val ' +
+            newPlaceholderName +
+            ' = ' +
+            tablePlaceholder +
+            '.get_column("' +
```
lars-reimann, line 89:
```suggestion
            '.getColumn("' +
```

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+            columnName +
+            '").missing_value_ratio(); \n'
+        );
+    }
+
+    private sdsStringForColumnNames(tableIdentifier: string, newPlaceholderName: string): string {
+        return 'val ' + newPlaceholderName + ' = ' + tableIdentifier + '.column_names; \n';
```
lars-reimann, line 96:
```suggestion
        return 'val ' + newPlaceholderName + ' = ' + tableIdentifier + '.columnNames; \n';
```

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    private sdsStringForIDnessByColumnName(columnName: string, tablePlaceholder: string, newPlaceholderName: string) {
+        return (
+            'val ' + newPlaceholderName + ' = ' + tablePlaceholder + '.get_column("' + columnName + '").idness(); \n'
```
lars-reimann, line 101:
```suggestion
            'val ' + newPlaceholderName + ' = ' + tablePlaceholder + '.getColumn("' + columnName + '").idness(); \n'
```

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+        columnName: string,
+        tablePlaceholder: string,
+        newPlaceholderName: string,
+    ) {
+        return (
+            'val ' +
+            newPlaceholderName +
+            ' = ' +
+            tablePlaceholder +
+            '.get_column("' +
```
lars-reimann, line 115:
```suggestion
            '.getColumn("' +
```

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,412 @@
...
+            columnName +
+            '").plot_histogram(); \n'
```
lars-reimann, line 117:
```suggestion
            '").plotHistogram(); \n'
```

packages/safe-ds-eda/src/components/profiling/ProfilingInfo.svelte
lars-reimann: > so I can also just change to "profiling image" for now

OK.

packages/safe-ds-eda/src/webviewState.ts
SmiteDeluxe: column[0] is for the index of order of the column, so what position it is, initially planned for restoring it after fetching a new table from the Runner through an action. When faced with the action history and restoring it (in the next issue) I was looking a lot more at this and will change how it works, so will be addressed there.

packages/safe-ds-eda/src/webviewState.ts
SmiteDeluxe: but then when I display the profiling in my HTML I will have to iterate through that Profiling info and can not anymore just access it by the column that I am iterating over anyway. The Svelte stores that I use (particularly the history) will be split up a bit more in the future, so that I can update only a part and not all HTML gets rerendered on say an addition to the history. So this will be addressed in another capacity, but afraid that in my opinion this profiling still belongs here with the column object and the state is not planned to get flatter itself, only the handling of it.

packages/safe-ds-eda/types/state.d.ts
SmiteDeluxe: Already addressing this in a new branch where I work out more details of the state and history

packages/safe-ds-eda/types/state.d.ts
SmiteDeluxe: This specifically defines possible filters though that I pass to my filter component and then based on this show those filters that are possible. Like `SearchStringFilter `defines the string searched for, while `PossibleSearchStringFilter `does not, and `ValueRangeFilter `defines the selected current min and max, while `PossibleValueRangeFilter `only defines the total possible min and max.
`ColumnFilter` is in theory `NumericalFilter | CategoricalFilter` but I use those directly depending on the column type and never so far unified.

SmiteDeluxe:
@lars-reimann @WinPlay02 I am confused as to how I can use those stubs. If I just try to use the functions like fromCsvFile on a table, I get an error if not in my own stub file.

lars-reimann:
Did you rebuild everything?

SmiteDeluxe:
> Did you rebuild everything (`npm run build` in project root)?

I apparently had to remove my `package a`  declaration that linked my pipeline file to my stub file. After that I can use the correct stubs.

SmiteDeluxe:
Now with the stubs from #950 the getColumns() Method just throws "unhashable type: 'Table'", so all my profiling that builds up on this does not work.
image

This is the code:
```
package b

from safeds.data.tabular.containers import Table

pipeline mainpipeline {
    val Tithhghgani = Table.fromCsvFile("/home/jonas/titanic.csv");
    val test = Tithhghgani.getColumn("Sex");
    val test2 = test.missingValueRatio();
    val Cereaddfkf5fl = Table.fromCsvFile("/home/jonas/cereal.csv");
}

pipeline secondary {
    val Titadssfsffjgfg3d832 = Table.fromCsvFile("/home/jonas/netflix_titles.csv");
    val Cereaf2gfgfff8dl = Table.fromCsvFile("/home/jonas/tweets.csv");
}
```

lars-reimann:
> Had to import manually.

Yep, having this as well. This is a bug: #974.

packages/safe-ds-eda/src/webviewState.ts
lars-reimann: How about storing the initial index in the column object (i.e. column[1]) then?

WinPlay02:
> Now with the stubs from #950 the getColumns() Method just throws "unhashable type: 'Table'", so all my profiling that builds up on this does not work. @lars-reimann @WinPlay02

@SmiteDeluxe I'd guess the memoization infrastructure is now (properly) used (as opposed to before) because the official stubs describe the (im)pureness of the functions correctly.

Thus, only the update to the latest (unreleased) library + runner will fix this (as described and fixed in https://github.com/Safe-DS/Library/pull/573.)

@lars-reimann In this context, I'd request a release for the runner + library including this fix.

lars-reimann:
@SmiteDeluxe @WinPlay02:

* New library and runner versions are released.
* #976 now requires the latest runner version.

SmiteDeluxe:
@lars-reimann @WinPlay02 Now the Table hashing seems to be working but I get another error for getColumn():
image

My code is still the same as above. My steps were: merge main of DSL into this branch, run `npm i `and `npm run langium:generate` and pull on Runner and run `poetry install`

SmiteDeluxe:
@lars-reimann @WinPlay02 Now the Table hashing appears to be working but the getColumn() command fails because of another reason:
image

My code is still:
```
package b

from safeds.data.tabular.containers import Table

pipeline mainpipeline {
    val Tithhghgani = Table.fromCsvFile("/home/jonas/titanic.csv");
    val test = Tithhghgani.getColumn("Sex");
    val test2 = test.missingValueRatio();
    val Cereaddfkf5fl = Table.fromCsvFile("/home/jonas/cereal.csv");
}

pipeline secondary {
    val Titadssfsffjgfg3d832 = Table.fromCsvFile("/home/jonas/netflix_titles.csv");
    val Cereaf2gfgfff8dl = Table.fromCsvFile("/home/jonas/tweets.csv");
}
```
And my steps were to merge main into this branch, run `npm i` and `npm run langium:generate` and then pull Runner and run `poetry install`

WinPlay02:
@SmiteDeluxe Could you attach the full log here? The messages before the error would be helpful to find the issue

packages/safe-ds-eda/src/webviewState.ts
SmiteDeluxe: My plan was to have all the internal actions (not requiring a runner call, like hide column, reorder etc.) be restored inside the webview.  Since internal actions will be stored in the history too, they may not need extra saving in the state so I could restore the column order from just the history and thus get rid of the index in the state object for columns. Then in the history object for reordering I will store the new order as a map of column name to index and whenever the table updates with a request to the Runner I will iterate over the table to put it in the right order of the last reorder history item before showing.

SmiteDeluxe:
@WinPlay02 Here you go
image

WinPlay02:
> @WinPlay02 Here you go image

@SmiteDeluxe I see, thanks. I'm looking into a fix

packages/safe-ds-vscode/src/extension/mainClient.ts
```diff
@@ -178,8 +180,59 @@ const registerVSCodeCommands = function (context: vscode.ExtensionContext) {
                         vscode.window.showErrorMessage('No .sdspipe file selected!');
                         return;
                     }
+
+                    // Getting of pipeline name
+                    const document = await getPipelineDocument(editor.document.uri);
+                    if (!document) {
+                        vscode.window.showErrorMessage('Internal error!');
```
lars-reimann, line 187: We should not "scream" at the user in error messages. Rather end with a full stop than with an exclamation point.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,415 @@
...
+        return new Promise(async (resolve, reject) => {
+            if (!this.baseDocument) {
+                reject('Document not found');
+                return;
+            }
+
+            const documentText = this.baseDocument.textDocument.getText();
+
+            // Find pattern "pipeline <pipelineName> {" and add the SDS code before the closing bracket of it
+            const pipelineStart = documentText.indexOf('pipeline ' + this.pipelineName);
```
lars-reimann, line 39: That's not going to work if there is extra whitespace or comments or any other hidden tokens between the `pipeline` keyword and the name. Generally, you should avoid parsing the document text again, that's what we have the parser created by Langium for. 

Instead, start with the AST node for the pipeline and get it's `$cstNdde`. From there, you can access the end position or offset of the pipeline, which is where you can add your code.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,415 @@
...
+            let currentMax = 0;
+            for (const [columnName, columnValues] of Object.entries(pythonTableColumns)) {
+                if (!Array.isArray(columnValues)) {
+                    continue;
+                }
+                if (currentMax < columnValues.length) {
+                    currentMax = columnValues.length;
+                }
+
+                const isNumerical = typeof columnValues[0] === 'number';
```
lars-reimann, line 175: Even if the first entry of a column is numeric, the entire column might still be categorical. For example, the column might contain values 0 for false and 1 for true.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,415 @@
...
+        const missingValueRatioMap = new Map<string, number>(); // Saved by random placeholder name
+
+        const columnNameToPlaceholderIDnessNameMap = new Map<string, string>(); // Mapping random placeholder name for IDness back to column name
+        const idnessMap = new Map<string, number>(); // Saved by random placeholder name
+
+        const columnNameToPlaceholderHistogramNameMap = new Map<string, string>(); // Mapping random placeholder name for histogram back to column name
+        const histogramMap = new Map<string, Base64Image | undefined>(); // Saved by random placeholder name
+
+        // Generate SDS code to get missing value ratio for each column
+        for (let i = 0; i < columns.length; i++) {
```
lars-reimann, line 214: Could be replaced by `for..of`

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,415 @@
...
+        const profiling: { columnName: string; profiling: Profiling }[] = [];
+        for (const column of columns) {
+            // Base info for the top of the profiling
+            const missingValuesRatio =
+                missingValueRatioMap.get(columnNameToPlaceholderMVNameMap.get(column[1].name)!)! * 100;
+
+            const validRatio: ProfilingDetailStatistical = {
+                type: 'numerical',
+                name: 'Valid',
+                value: missingValuesRatio ? (100 - missingValuesRatio).toFixed(2) + '%' : '100%',
```
lars-reimann, line 292: Keep the ratio here. Do the conversion to a string (representation) in the Svelte component. Now that the profiling data is no longer only separated into top & bottom, this is possible.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -0,0 +1,415 @@
...
+                interpretation: 'good',
+            };
+
+            const missingRatio: ProfilingDetailStatistical = {
+                type: 'numerical',
+                name: 'Missing',
+                value: missingValuesRatio ? missingValuesRatio.toFixed(2) + '%' : '0%',
```
lars-reimann, line 299: Keep the ratio here. Do the conversion to a string (representation) in the Svelte component.

packages/safe-ds-eda/src/components/TableView.svelte
```diff
@@ -3,55 +3,87 @@
...
 
-        // Use the logarithm of the character count, and scale it
-        const width = baseWidth + Math.log(columnName.length + 1) * scale;
+            // const baseWidth = 35; // Minimum width
+            // const scale = 55;
 
-        // Save the width for future use
-        savedColumnWidths.set(columnName, width);
+            // // Use the logarithm of the character count, and scale it
+            // const width = baseWidth + Math.log(columnName.length + 1) * scale;
```
lars-reimann, line 68: Still needed?

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
SmiteDeluxe: How to get this type then. Can it also be passed with the table somehow from Runner?

SmiteDeluxe:
@WinPlay02 Now the `getColumn()` call on it's own works, just like `missingValueRatio()`, the problem is that a chained call of `.getColumn("Sex").missingValueRatio()` does not work and throws the error below:
image

code:
```
package b

from safeds.data.tabular.containers import Table
from safeds.data.tabular.containers import Column

pipeline mainpipeline {
    val Titanghijjihfgk = Table.fromCsvFile("/home/jonas/titanic.csv");
    val test = Titanghijjihfgk.getColumn("Sex").missingValueRatio();
    val Cereaddfkf5fl = Table.fromCsvFile("/home/jonas/cereal.csv");
}

pipeline secondary {
    val Titadssfsffjgfg3d832 = Table.fromCsvFile("/home/jonas/netflix_titles.csv");
    val Cereaf2gfgfff8dl = Table.fromCsvFile("/home/jonas/tweets.csv");
}
```

This worked before with my own fake stubs.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
SmiteDeluxe: So it would need extra calls, I see. What would I use the idness for in this context? To display a column with just 0 and 1 as categorical you mean? I would personally just use the types the columns actually have in the background and not interpret that myself

WinPlay02:
> @WinPlay02 Now the `getColumn()` call on it's own works, just like `missingValueRatio()`, the problem is that a chained call of `.getColumn("Sex").missingValueRatio()` does not work and throws the error below: image
> 
> code:
> 
> ```
> package b
> 
> from safeds.data.tabular.containers import Table
> from safeds.data.tabular.containers import Column
> 
> pipeline mainpipeline {
>     val Titanghijjihfgk = Table.fromCsvFile("/home/jonas/titanic.csv");
>     val test = Titanghijjihfgk.getColumn("Sex").missingValueRatio();
>     val Cereaddfkf5fl = Table.fromCsvFile("/home/jonas/cereal.csv");
> }
> 
> pipeline secondary {
>     val Titadssfsffjgfg3d832 = Table.fromCsvFile("/home/jonas/netflix_titles.csv");
>     val Cereaf2gfgfff8dl = Table.fromCsvFile("/home/jonas/tweets.csv");
> }
> ```
> 
> This worked before with my own fake stubs.

@SmiteDeluxe A fix is prepared in #987

SmiteDeluxe:
@WinPlay02 @lars-reimann profiling is working again now!

lars-reimann:
@SmiteDeluxe For me, profiling is stuck on `Loading...`. Do I need to do anything special after clicking `Explore Table`?

lars-reimann (approved):
Congratulations on finishing this. It looks great and works well.

PR #982, opened 2024-04-03
fix: generation of memoized class member calls

WinPlay02:
- fix: generating memoized class member calls (on non-static members) should still use the class for python code generation
- added testcase
------------------
Fixes the (last) issue in #954


packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -1083,6 +1090,10 @@ export class SafeDsPythonGenerator {
         throw new Error('Callable of provided call does not exist or is not a declaration.');
     }
 
+    private getClassQualifiedNameForMember(expression: SdsClassMember): string {
+        return `${this.getPythonNameOrDefault(AstUtils.getContainerOfType(expression.$container, isSdsDeclaration)!)}.${this.getPythonNameOrDefault(expression)}`;
```
lars-reimann, line 1094: Classes can be nested, which is not handled here:

```
class Outer {
    class Nested {
        @Pure
        static fun f() -> result: Boolean
    }
}
```

lars-reimann (changes requested):

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
WinPlay02: If understood correctly, this should now be fixed

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -1083,6 +1090,16 @@ export class SafeDsPythonGenerator {
         throw new Error('Callable of provided call does not exist or is not a declaration.');
     }
 
+    private getClassQualifiedNameForMember(expression: SdsClassMember): string {
+        const classMemberPath = [this.getPythonNameOrDefault(expression)];
+        let enclosingClass = AstUtils.getContainerOfType(expression.$container, isSdsClass);
+        while (enclosingClass) {
+            classMemberPath.push(this.getPythonNameOrDefault(enclosingClass));
```
lars-reimann, line 1097: Could also `unshift` here to insert at the start. Then you don't need to reverse at the end.

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -1083,6 +1090,16 @@ export class SafeDsPythonGenerator {
         throw new Error('Callable of provided call does not exist or is not a declaration.');
     }
 
+    private getClassQualifiedNameForMember(expression: SdsClassMember): string {
```
lars-reimann, line 1093:
```suggestion
    private getClassQualifiedNameForMember(node: SdsClassMember): string {
```

packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -1083,6 +1090,16 @@ export class SafeDsPythonGenerator {
...
+        const classMemberPath = [this.getPythonNameOrDefault(expression)];
+        let enclosingClass = AstUtils.getContainerOfType(expression.$container, isSdsClass);
+        while (enclosingClass) {
```
lars-reimann: Last refactoring:

```suggestion
        const classMemberPath = [];
        let current = node;
        while (enclosingClass) {
```

Then the logic to update `current` is only needed once.

lars-reimann (approved):
Excellent, thank you.

PR #983, opened 2024-04-04
fix: correctly import declarations for member functions

WinPlay02:
- fix: correctly import declarations for member functions
- replace import functionality


packages/safe-ds-lang/src/language/generation/safe-ds-python-generator.ts
```diff
@@ -1020,7 +1020,12 @@ export class SafeDsPythonGenerator {
         );
         const fullyQualifiedTargetName = this.generateFullyQualifiedFunctionName(expression);
         if (!containsOptionalArgs && isSdsMemberAccess(expression.receiver)) {
-            const referenceImport = this.createImportDataForReference(expression.receiver.member!);
+            const classDeclaration = this.getOutermostContainerOfType(callable, isSdsClass)!;
+            const referenceImport = this.createImportDataForNode(
+                classDeclaration,
+                expression.receiver.member!,
+                classDeclaration.name,
```
WinPlay02, line 1027: I'm still a bit doubtful about the `refName` (3rd parameter) here, but considering that the user has no way of specifying it here, that should be fine.

lars-reimann: Maybe we can make the 3rd parameter optional, with `declaration.name` as the default?

lars-reimann (approved):
Looks good. Code like

```
package test

pipeline myPipeline {
    val titanic = Table.fromCsvFile("C:/Users/Lars/OneDrive/Desktop/test/titanic.csv");
    val age = titanic.getColumn("age");
    val idness = age.idness();
    val string = idness.toString();
}
```

can now be executed.

PR #1010, opened 2024-04-08
feat: trigger EDA tool via code lenses

lars-reimann:


lars-reimann:
@SmiteDeluxe What do you think of this? Could this fully replace the `runEdaFromContext` command and entries in the context menu? Compared to those, it really is context-dependent and only shown for placeholders with type `Table`.

SmiteDeluxe:
@lars-reimann I know I am a bit late on this but I agree this is great! I was actually thinking about something similar to talk to you about, but guess that is now resolved! And I figure we might still want the context menu one in case of bugs/changes in vscode and it's sometimes breaking updates (in my experience) and as following the UX standard of allowing multiple paths to a goal.

PR #1019, opened 2024-04-10
feat: support relative paths

lars-reimann:
### Summary of Changes

Use the runner functionality added in https://github.com/Safe-DS/Runner/pull/83 to support relative paths in pipeline files.


WinPlay02 (approved):
Manually verified, everything seems to work expected

lars-reimann:
Nice, thanks for checking.

PR #1094, opened 2024-04-24
fix: runner not shutting down when extension was closed

lars-reimann:
### Summary of Changes

Graceful shutdown of the runner now works again.


lars-reimann:
Since the `shutdown` message was not sent, we got into the 2.5s timeout. Apparently, during this time, the extension host was killed, so it could never tree-kill the spawned processes.

Issue #1079, opened 2024-04-22
"Run until" code generation for list of placeholders


lars-reimann:
### Is your feature request related to a problem?

Currently, the code generator can take a single target placeholder. It would be more useful, if it would take a list of target placeholders instead, e.g. for @SmiteDeluxe's EDA tool, which currently fully executes the pipeline.

It would also be nice, if the runner would immediately send the values of the target placeholder back.

### Desired solution

* Update `getStatementsNeededForPartialExecution` in the code generator
* For target placeholders, generate a call `send_placeholder_value` (needs runner update). This should immediately send the placeholder value.
* For other placeholder, generate a call `report_progress` (needs runner update). This just sends the name of the placeholder.

PR #1114, opened 2024-04-27
feat: partial code generation for multiple targets

lars-reimann:
Closes partially #1079

### Summary of Changes

Partial code generation now works for multiple targets instead of just one.


lars-reimann:
@SmiteDeluxe You can now pass a list containing the names of your target placeholders to  `services.runtime.Runner.executePipeline`. Then you can just dump your additional code into the pipeline and don't need to change the existing code.

PR #1170, opened 2024-05-14
refactor: clean up state and separate it

### LINKED ISSUE #1169: Improve EDA state

Messy state where any update to any part updates state as a whole. Some parts like column widths are kept in a store inside the components, while others are part of the big state object.

### Desired solution

More separated into individual components and all the ones needed across components in the webviewstate.ts while the ones only needed in one defined and managed there. This will make the typing also cleaner and the extension will not have to insert placeholder data it should not really manage anyway like visible vs. total rows. Only possible because the state is not saved to vscodes global state anymore.

_No response_

### Screenshots (optional)

_No response_

_No response_

SmiteDeluxe:
Closes #1169 

### Summary of Changes

- Split up the currentState into tabs, table and history stores, as no saving to vscode global state anymore so not needed to save as one big object
- This means that svelte does not have to reevaluate so many things when something small changes, as in a new history item does not cause update to whole state anymore that causes table to update in background
- Also easier to manage and update
- tableIdentifier in table object, defaultState property not needed anymore and UserSettings not yet
- columns don't have a number anymore as order is just array order
- table filter is object that can have all the tablefilters but not multiple of one type
- columnWidths still shall remain own store in component, as otherwise resize drag would cause massive amount of updates to the table store
- setCurrentState msg is not setInitialTable and takes Table, error if executed more than once


lars-reimann:
Does this PR add new features or is it "only" to refactor? Just asking because of the PR title.

SmiteDeluxe:
> Does this PR add new features or is it "only" to refactor? Just asking because of the PR title.

Oops sorry, not intended. Yes only refactor, changed that title.

lars-reimann (approved):
I've found no regressions and appreciate the reduced nesting. Thank you.

PR #1181, opened 2024-05-27
feat: hide columns in EDA

### LINKED ISSUE #1171: Hide column feature in EDA

Lack of overview with lots of columns & messy visualizations.

### Desired solution

Hiding a column so that it is minimized + an indication/interface where it lists hidden cols to show them again, possibly bottom of sidebar and move row count to there as well.
In visualizing calls these will be passed to make sure they are excluded. In manipulating ones they will still be included to also apply manipulative actions on them so that any current filters/sorts also apply to them on re-showing them.

_No response_

### Screenshots (optional)

_No response_

_No response_

SmiteDeluxe:
Closes #1171 

### Summary of Changes

- You can now hide columns by right clicking them
- Showing either by right click again or by clicking on new column count in sidebar footer (row count there too now) 
- Hidden columns excluded from visualizations that affect whole table (heatmap)
- Still possible to select in other visualizations but marked red and with "hidden" in case you still want to use that column
- Hiding/Showing cols marks whole table vis as "outdated"
- TabContent evaluates which vis types allow non numeric columns and manages buildATab and drop downs based on that

Other fixes/refactors:
- Context menu on columns
- Profiling errors present evaluation
- Color css vars
- Sidebar width reactivity


packages/safe-ds-eda/src/apis/historyApi.ts
```diff
@@ -220,3 +227,30 @@ const evaluateMessagesWaitingForTurn = function () {
...
+                        return {
+                            ...t,
+                            outdated: true,
+                        };
+                    } else {
+                        return t;
+                    }
+                } else {
+                    return t;
+                }
```
lars-reimann, line 250:
```suggestion
                if (t.type !== 'empty' &&
                    t.columnNumber === 'none' &&
                    get(table)?.columns.find((c) => c.name === entry.columnName)?.type === 'numerical') {
                        // UPDATE the if in case there are none column tabs that do not depend on numerical columns
                    return {
                        ...t,
                        outdated: true,
                    };
                } else {
                    return t;
                }
```

packages/safe-ds-eda/src/apis/historyApi.ts
```diff
@@ -198,6 +202,9 @@ const deployResult = function (result: RunnerExecutionResultMessage) {
         tab.id = crypto.randomUUID();
         tabs.update((state) => state.concat(tab));
         currentTabIndex.set(get(tabs).indexOf(tab));
+    } else if (resultContent.type === 'table') {
+        updateTabOutdated(historyEntry);
+        throw new Error('Not implemented');
```
lars-reimann, line 207: What's the planned behavior in this case?

packages/safe-ds-eda/src/components/ColumnCounts.svelte
```diff
@@ -0,0 +1,146 @@
...
+
+    const showColumn = function (columnName: string) {
+        table.update(($table) => {
+            return {
+                ...$table!,
+                columns: $table!.columns.map((column) => {
+                    if (columnName === column.name) {
+                        return {
+                            ...column,
+                            hidden: !column.hidden,
```
lars-reimann, line 52:
```suggestion
                            hidden: false,
```

Since the function is called "showColumn" rather than "toggleColumnVisibility", I find it clearer to be explicit here.

packages/safe-ds-vscode/src/extension/eda/apis/runnerApi.ts
```diff
@@ -269,6 +275,18 @@ export class RunnerApi {
...
+
+    private sdsStringForRemoveColumns(columnNames: string[], tablePlaceholder: string, newPlaceholderName: string) {
+        return (
+            'val ' +
+            newPlaceholderName +
+            ' = ' +
+            tablePlaceholder +
+            '.removeColumns(["' +
+            columnNames.join('","') +
+            '"]); \n'
```
lars-reimann, line 287: This would break if the function gets called with an empty list for `columnNames`. It's safer to first map over the `columnNames` to enclose them in double quotes, and then join them together.

packages/safe-ds-eda/src/components/tabs/TabContent.svelte
```diff
@@ -130,6 +158,33 @@
                 });
             }
         }
+
+        // Check if columns now require numerical column only and if so check if selected column(s) are numerical
```
lars-reimann, line 162: Unclear what's meant by "columns [that] require numerical column".

packages/safe-ds-eda/src/components/tabs/TabContent.svelte
```diff
@@ -11,7 +11,34 @@
     export let tab: Tab | EmptyTab;
     export let sidebarWidth: number;
 
-    const columnNames = $table?.columns.map((column) => column.name) || [];
+    const allowsNonNumericalColumns = function (type: string) {
+        return type === 'histogram';
+    };
+
+    const columnNames = derived(
```
lars-reimann, line 18: Contains more than just column names. Needs a better name, maybe "columnOptions".

packages/safe-ds-eda/src/components/tabs/TabContent.svelte
```diff
@@ -11,7 +11,34 @@
...
+        ($table) =>
+            $table?.columns.map((column) => {
+                return {
+                    name: column.name,
+                    color: column.hidden ? 'var(--error-color)' : undefined,
+                    comment: column.hidden ? 'hidden' : undefined,
+                };
+            }) || [],
+    );
+    const numericalColumnNames = derived(
```
lars-reimann, line 29: Contains more than just column names. Needs a better name, maybe "numericalColumnOptions".

lars-reimann (changes requested):
Looks awesome, as usual. Just a few stylistic requests.

packages/safe-ds-eda/src/apis/historyApi.ts
SmiteDeluxe: @lars-reimann see #1183 where this is implemented. Essentially update column values and column types and then all other changes based on the type of action (like set respective column to sorted if it was a sort).

lars-reimann (approved):

Issue #1031, opened 2024-04-12
Getting Started section & tutorials in documentation


lars-reimann:
To help people get started with development in Safe-DS, we need a getting started section in the documentation and tutorials for recurring tasks.

PR #1202, opened 2024-06-07
docs: first regression program

Muellersen:
Closes partially #1031

### Summary of Changes

This pull request adds a new page to the getting started tab. This page displays the first steps one has to follow, in order to write their first safe ds program, that implements a regression model. It also offers a section that shows the full code, so that one can copy and paste for convenience.


lars-reimann (approved):
Nicely explained, thank you.

PR #1213, opened 2024-06-21
fix: eda table right click improvements

### LINKED ISSUE #1180: EDA: table right click improvements

Unintuitive that:
- No row right clicks
- Two column plotting by selection while one column by current right clicked column
- No right clicks on table cells

### Desired solution

- Row right clicks for selections
- Submenus in column right click:
     - "Plot Selection" if one or two selected => plot from selected columns
     - "Plot Column" that plots one column plots on current right clicked columns
- "Plot Selection" also possible if right click on cells

_No response_

### Screenshots (optional)

_No response_

_No response_

SmiteDeluxe:
Closes #1180 

### Summary of Changes
(Includes History PR changes)

- Submenus in column right click:
    - "From Selection" if one or two selected => plot from selected columns or hide from selection
    - "From Click" that plots one column plots on current right clicked columns and hides column
- "From Selection" also possible if right click on cells
- All plotting now checks for column type where only numerical is allowed and shows greyed out with tooltip incl. reason if not possible
- Right clicks if preventClicks is true only allowed if on same element type, so can change position of same right click by doing it again
- With hiding mult cols from selection the toggleHideColumn method now takes array and only does it if all columns have same hidden state; for now adds separate history entries as creating a new action type for this would be too much


lars-reimann (changes requested):
The context-menu on the header of the `ids` column no longer works for the following program:

```
package demo

pipeline whoSurvived {
    val a = Table({"ids": [1, 2, 3]});
}
```

`Plot Histogram` and `Plot Boxplot` no longer do anything. It still seems to work for tables with multiple columns.

SmiteDeluxe:
@lars-reimann fixed! That annoying thing of !0 evaluating to true ...

lars-reimann (approved):
👍

PR #1306, opened 2025-01-04
ci: use default build steps of ReadTheDocs again

lars-reimann:
Closes #XYZ


lars-reimann:
Closed since our current setup works, but this new one does not.

